PR #22, opened 2021-05-20
Mlintegrate

rohanakut:
The code contains the docker file and my python code. You can test it by creating a json request. the code will return a json. In the URL make sure to pass the URL link of the image and not the webpage as that will break the code and return a 404 error


JRegimbal (changes requested):
The biggest problem is this pull request would delete every other file in the branch! It should only add your code to its own subdirectory under "docker". Aside from that, there are some formatting issues that impede readability, some possible bugs, and issues that would prevent it from working in practice. See the comments for details.

PR #23, opened 2021-05-21
Mlintegrate

rohanakut:
My code is mentioned in machine_learning folder


JRegimbal (changes requested):
Almost there, but the main issue is that you deleted the schemas submodule! We use that to keep a reference to a specific version of the schemas branch and update that as needed without file inconsistency. Replacing it with regular files makes it easy to break.

Also as a general note, please use smaller commits as discussed [in this blog](https://www.perforce.com/blog/vcs/git-best-practices-git-commit). Large commits like this make seeing what was done and fixing problems very difficult. For example if there was a commit where you deleted the submodule and nothing else, we could then revert that commit easily.

PR #38, opened 2021-05-27
Mlintegrate

rohanakut:


JRegimbal (changes requested):
I've suggested some changes here, but I'd also like you to run this though autopep8 or a similar tool so it follows PEP-8 code standards. After that I can take another look. And again you can just push the changes to this branch and request another review in this same pull request; that'll improve readability of this process later on since it'll all be in one place.

JRegimbal (changes requested):
Getting there! Just a few things remaining.

PR #45, opened 2021-06-10
Added scene preprocessor

rohanakut:


JRegimbal (changes requested):
You didn't include a .gitignore and so there are many files that should not have been tracked. You're also missing comments and included untracked schema files in this commit again. There should be more organization of the files you use to actually run this, perhaps in subfolders. We should also set up Git LFS for the model file since it is very large.

JRegimbal (changes requested):
There still are files that should be deleted like .DS_Store and the cache files. Please add a gitignore for python to this.

JRegimbal (changes requested):

JRegimbal (approved):

PR #47, opened 2021-06-17
Added grouping schema

rohanakut:


JRegimbal (changes requested):
Fix formatting on the JSON and remove the .DS_STORE first.

JRegimbal (changes requested):
Add documentation of what should actually go in here using the description tags.

JRegimbal (changes requested):

JRegimbal (changes requested):

rohanakut (commented):
The array of array of objects is chosen because according to my logic an array of object would hold a list of IDs which have same type. Hence an array of array of objects would then hold multiple such arrays and thereby support grouping of multiple objects. If i were to consider just an array of objects then it would allow me to do just 1 grouping

JRegimbal (approved):
Looks good. Verify that the structure we discussed passes validation against it and then you can merge it.

PR #48, opened 2021-06-25
Added segmentation schema

rohanakut:


JRegimbal (changes requested):
Lots of problems here, not least of which is that this isn't even valid JSON. Fix that big stuff, correct the typos, and check out the other comments.

JRegimbal (changes requested):

JRegimbal (changes requested):
Specifications for centroid and area mostly fine, but specify 2 values only for centroid and also add useful description fields for each since these files are currently the only documentation we have. Please also see the other comments.

JRegimbal (approved):
Seems good for right now so I'll merge. Heads up though: ID names are changing and it turns out how we've been doing some array stuff isn't official for JSON schema draft 2020-12. This will all be fixed in a later PR, but is fine for right now provided it works with your code.

PR #50, opened 2021-06-28
Added grouping preprocessor

rohanakut:


JRegimbal (changes requested):
No .DS_Store files

JRegimbal (commented):
This is fine, though you should also exclude them from commits in your local setup since (AFAIK) there is no case where you'd want to commit .DS_Store. I'll take a look at this for correctness in a bit.

PR #112, opened 2021-09-11
Added categoriser schema

rohanakut:
The pull request contains categoriser schema.


JRegimbal (changes requested):
Good, but ideally there would be a finite number of options that "category" can take.

JRegimbal (approved):
Alright this looks good to me!

PR #132, opened 2021-10-07
Pr for issue 79

rohanakut:
The PR is created for I[ssue 79](https://github.com/Shared-Reality-Lab/auditory-haptic-graphics-server/issues/79). The entire docker repo has been tested on unicorn and the semantic segmentation model works as mentioned in the issue. 

I have commented out the object detection code in docker-compose.yml because unicorn gives an OOM error if both models are run simultaneously.


JRegimbal (changes requested):
Please remove the commenting out of object-detection

Issue #121, opened 2021-09-20
Create preprocessor that takes embedded map lat/lng request, gets Autour data, produces JSON


jeffbl:
#71 is to investigate what information we can obtain from an embedded google map, but @Cybernide indicated last week that it looks like lat/lng is accessible. This work item is a followup from our discussion last Wednesday morning, to accept the information from Shared-Reality-Lab/auditory-haptic-graphics-browser#47 and create a preprocessor that takes the lat/lng, queries the Autour server, and produces JSON output for the current position. Note that this information should include the lat/lng of the query as well, so that the handler can identify the most central POI as being the main POI for the map itself. (This may need more fleshing out, since how exactly we know the exact POI the embedded map is referring to partially depends on what we can extract from #71...)

This will require working with @florian-grond to make sure the JSON is usable for creating the handler that generates the audio experience.

PR #145, opened 2021-10-21
Create schema for preprocessor

aidanwilliams09:
Referencing Issue #121 schemas submodule

New schema created for autour data

### DIFF
--- a/preprocessors/autour.schema.json
+++ b/preprocessors/autour.schema.json
@@ -32,7 +32,7 @@
             "pattern": "^https?://.*"
         },
         "places": {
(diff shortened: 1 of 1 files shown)


JRegimbal (changes requested):
Overall a really good pull request! Just need to be a bit nitpicky on some lat long specifications. You may want to make a definition for lat/lon and then reference it when necessary for consistency.

JRegimbal (changes requested):
An "object" cannot have items.

JRegimbal (approved):

PR #151, opened 2021-10-27
Added condition to run only if chart is present

rohanakut:
I have added the condition for running this preprocessor only when the chart image is supplied. The code is working and tested on unicorn

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,20 +50,20 @@ services:
             - ./config/express-common.env
         labels:
             ca.mcgill.a11y.image.handler: enable
-    object-detection:
-        build:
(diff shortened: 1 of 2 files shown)


JRegimbal (changes requested):
There's no guarantee first categoriser will be running in every setup and, if it is, the check should be that it is not a chart.

JRegimbal (changes requested):
Don't put the same code in two locations. If we ever need to do maintenance it'll leave us open to inconsistent behavior between branches that should act the same. Either restructure the logic or break parts of this out into a separate function.

JRegimbal (approved):
Looks good to me. Have you tested on Unicorn?

PR #258, opened 2022-01-06
Autour-style POI rendering

### LINKED ISSUE #122: Create handler to render Autour audio for a particular lat/lng
Once #121 produces the JSON for POIs around a specific lat/lng via the Autour server, this work item is to create the handler that generates the audio rendering. This could be as simple as re-creating one pass around the user, as they would hear on the smartphone version of Autour, but may could also later add:

- Playing the actual location of interest from the embedded map first, or in a special way
- Autour-style controls to get details on locations
- Other renderings, such as the category name and then spatialized "pings" for each match around the user, like the current object recognizer

Not sure if it is @Sabrina-Knappe or @Cybernide who will weigh in on the associated UI issues, but when decided, can be added as asisgnee to this work item.

JRegimbal:
This resolves #122 by using earcons and TTS to present the points-of-interest around a particular lat/lon sent in a request to the server.

At most twenty points of interest (by those closest) are displayed with earcons generated from the category number returned by autour. The mapping of earcon to category is arbitrary. POIs are then ordered in a sweep starting from due north and moving clockwise in a sweep. This has been tested in a local pipeline without using the browser, but instead using a handcrafted request that complies with the current requirement of the autour preprocessor to use the `url` key (see #285 for more information). It works with a variety of locations and points seem to be ordered correctly.

Note that the TTS for the place names is time consuming since they are less likely to have been already generated and cached.

## Required Information

## Pull Request Checklist

- [x] The Docker image builds


JRegimbal:
Currently stalled on testing since the output from the autour-preprocessor can't be generated right now and may significantly change.

JRegimbal:
Assigning to @florian-grond and @Cybernide for feedback on the example renderings as sent over Slack.

florian-grond:
This is great @JRegimbal, thanks for implementing the handler! maybe we can briefly touch base to discuss how we can shorten the earcon so as to give it a bit less "air time"

Cybernide:
Is there any reason why both pitch and spatialization signify direction? I was confused by that. Is this to help people who may have hearing loss in one or both ears?

florian-grond:
in spatial audio elevation cues are typically not as string as lateral cues, which is why we might want to add north-south to another salient dimension

JRegimbal:
@jeffbl @Cybernide let me know when this can be merged for at least testing on Unicorn.

jeffbl:
@JRegimbal apologies if you were waiting on me, but yes, please deploy to unicorn.

Issue #263, opened 2022-01-11
Move code snippet acknowledgements to where the actual code is


jeffbl:
In a bunch of files, there are acknowledgements of small snippets of code, which is the right thing to do. However, these are at the top of the files, and as soon as the code changes, the "line number" references become out of date. For example, the reference to line 19 in this file refers to itself, I assume because this was inserted before the code header was put into place:

https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/preprocessors/grouping/grouping.py

To make these references maintainable into the future, please move these acknowledgements directly to the place in the code where the snippet exists.

Ping to @JRegimbal in case documentation/examples should call this out specifically for future developer reference.

PR #264, opened 2022-01-11
Change citation position

rohanakut:
This PR is wrt to changes mentioned in issue [263](https://github.com/Shared-Reality-Lab/IMAGE-server/issues/263). The citation positions have been changed as requested in the issue. 

It is to be noted that the code has not been tested before pushing the changes as it involved only changing of the comments

### DIFF
--- a/preprocessors/categoriser/categoriser.py
+++ b/preprocessors/categoriser/categoriser.py
@@ -13,11 +13,6 @@
 # and our Additional Terms along with this program.
 # If not, see
 # <https://github.com/Shared-Reality-Lab/IMAGE-server/LICENSE>.
-# Lines 51-56 refered from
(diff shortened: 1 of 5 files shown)


jaydeepsingh25:
@rohanakut This PR also contains an new file preprocessors/sorting.schema.json. I guess you should remove that file. Or is it required in this commit??

rohanakut:
@jaydeepsingh25 The  sorting.schema.json is already present in the schemas folder. The reason that file is reflected in the current PR is because I am pointing to the latest commit of the `schemas `submodule. However, the `main` branch is pointing to an older commit. Due to this, the PR is suggesting that a change has been made to the schemas folder. I believe you would just need to repoint the `main` branch to the latest `schemas` commit and this change would disappear from this PR.

jaydeepsingh25 (approved):

PR #269, opened 2022-01-12
change 'Places' from array -> object to to conform to dictionary structure

BenMacnaughton:

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/preprocessors/autour.schema.json
+++ b/preprocessors/autour.schema.json
@@ -35,49 +35,45 @@
             "type": "object",
             "title": "Places",
             "description": "The places on the map",
-            "properties": {
-                "id": {
-                    "type": "string",
-                    "title": "Name",
-                    "description": "The name of the place",
-                    "minLength": 1
-                },
-                "ll": {
-                    "type": "array",
-                    "title": "Latitude/Longitude",
-                    "description": "The latitude and longitude of the place",
-                    "items": {
-                        "type": "number"
+            "additionalProperties": {
+                "type": "object",
+                "properties": {
+                    "ll": {
+                        "type": "array",
+                        "title": "Latitude/Longitude",
+                        "description": "The latitude and longitude of the place",
+                        "items": {
+                            "type": "number"
+                        }
+                    },
+                    "cat": {
+                        "type": "number",
+                        "title": "Category",
+                        "description": "The category of a place"
+                    },
+                    "title": {
+                        "type": "string",
+                        "title": "Title",
+                        "description": "The name of the place"
+                    },
+                    "from": {
+                        "type": "string",
+                        "title": "From",
+                        "description": "The server that returned the data of this place"
+                    },
+                    "addr": {
+                        "type": "string",
+                        "title": "Address",
+                        "description": "The address of the place"
+                    },
+                    "website": {
+                        "type": "string",
+                        "title": "Website",
+                        "description": "The website of the place"
                     }
                 },
-                "cat": {
-                    "type": "number",
-                    "title": "Category",
-                    "description": "The category of a place"
-                },
-                "title": {
-                    "type": "string",
-                    "title": "Title",
-                    "description": "The name of the place"
-                },
-                "from": {
-                    "type": "string",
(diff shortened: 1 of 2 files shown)


JRegimbal (changes requested):
The schema with the changes here doesn't make sense. You have places as type object, which means it should have properties, but instead assign it items, which imply an array. Can you provide a sample of what places should look like in the output?

BenMacnaughton:
Places appears in this format: `"places":{"A19RWFJe":{"addr":"1655 Rue Saint-Denis","cat":12,"from":"foursquareexplore","hours":"    ","ll":[45.515116,-73.563492],"phone":"5144395535","rating":8.2,"title":"Mâche!","website":"http://restaurantmache.com"},"A1lWUVQC":{"addr":"800 boul. De Maisonneuve Est, 17e étage","cat":71,"from":"foursquaresearch","ll":[45.516391999999996,-73.559775],"title":"Arrondissement Ville-Marie","website":"http://ville.montreal.qc.ca/villemarie"}}`

It is an unordered set of key-value pairs with the key being a unique place ID. We thought this would be a good way to track specific places down the line. We could change it to an array in the preprocessor but we felt that this K/V format could make it easier to identify places.

BenMacnaughton:
Two changes have been made: 1) Autour schema updated to reflect its dictionary structure using `additionalProperties` 2) a new maps request option has been created that includes a `placeID` field in place of the coordinates for situations where coordinates were not found on the browser side.

preprocessors/autour.schema.json
```diff
@@ -32,20 +32,12 @@
             "pattern": "^https?://.*"
         },
         "places": {
-            "type": "array",
+            "type": "object",
             "title": "Places",
             "description": "The places on the map",
-            "items": {
+            "additionalProperties": {
```
JRegimbal, line 38: This isn't how additionalProperties works...it is either true or false to indicate if you can have additional properties in an object along with those explicitly listed in properties.

request.schema.json
```diff
@@ -39,6 +39,10 @@
             },
             "required": [ "latitude", "longitude" ]
         },
+        "placeID": {
```
JRegimbal, line 42: What actually is this ID? Where does it come from? Why is it required?

JRegimbal (changes requested):
Schema still doesn't make sense...I'm also not sure why an object containing POIs indexed by some ID key is useful. Is there even a case where someone would know the ID ahead of time? To me, it seems like an array of objects where the object contains an ID field (as it was previously) would be better, though I'm willing to be convinced otherwise.

request.schema.json
aidanwilliams09: We haven't gotten to the browser side for this yet, but it is the substitute for a lat/long. eg the map may be a query of the Eiffel Tower instead of a lat/long and therefore we would make ```placeID='Eiffel+Tower,Paris+France```

JRegimbal: So this is a google specific ID? Or is it just a query?

BenMacnaughton: It is called a placeID by Google and can be structured in plain text or by a unique code as I understand

BenMacnaughton: The embedded map request allows for an english like query or an ID (https://developers.google.com/maps/documentation/places/web-service/place-id)

JRegimbal:
@BenMacnaughton @aidanwilliams09 what's the status of this?

BenMacnaughton:
@JRegimbal just waiting for approval I believe

JRegimbal:
Alright, you never reassigned so I was waiting on you. Is the autour preprocessor currently producing outputs that match this new schema?

request.schema.json
```diff
@@ -39,6 +39,10 @@
...
+            "description": "The place ID of the location identified in a map.",
```
JRegimbal, line 43:
```suggestion
            "description": "The Google place ID of the location identified in a map.",
```

JRegimbal (approved):

PR #270, opened 2022-01-13
Image server212

aidanwilliams09:

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,6 +49,8 @@ services:
             context: ./preprocessors/autour
             dockerfile: Dockerfile
         image: "autour:latest"
+        env_file:
+            - ./config/maps.env
         labels:
             ca.mcgill.a11y.image.preprocessor: 2
             ca.mcgill.a11y.image.port: 5000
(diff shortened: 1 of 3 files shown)


aidanwilliams09:
This pull request expands the maps module to allow for google places API calls in order to resolve unknown lat/long.

Additionally, this contains some basic refactors to reduce redundancy of validating data/schemas.

docker-compose.yml
```diff
@@ -49,6 +49,8 @@ services:
             context: ./preprocessors/autour
             dockerfile: Dockerfile
         image: "autour:latest"
+        env_file:
+            - ./config/maps.env
```
JRegimbal, line 53: Does this contain an API key?

docker-compose.yml
```diff
@@ -163,17 +165,6 @@ services:
             ca.mcgill.a11y.image.handler: enable
         volumes:
             - sc-store:/tmp/sc-store
-    generic-tts-handler:
```
JRegimbal, line 166: Why did you delete this?

JRegimbal (commented):
Some questions, but I don't review PRs for preprocessors, @jaydeepsingh25 does. Adding him here...

docker-compose.yml
aidanwilliams09: Yes, it contains Jeremy's Google API key, obviously this file was not added to the repo.

docker-compose.yml
aidanwilliams09: If you look at the lines above, there is a carbon copy of the ```generic-tts-handler```. I was not able to build the container when there were two processes named the same thing. If there is some purpose for there being two then I can add it back in.

JRegimbal: Oh no in that case don't worry. It only shows me the changes by default so it looked like you were deleting a whole thing...weird because we aren't having this problem in main as far as I can tell!

aidanwilliams09: See lines 157-167. 

I don't know how to link it because it's not part of the diff so it isn't giving me an option to comment on the lines.

aidanwilliams09:
Tested on unicorn and received back valid responses for query with lat/long and without. Additionally, tested bad query and received error codes and messages.

jaydeepsingh25 (approved):

PR #279, opened 2022-01-18
Updated file versions

rohanakut:
In order to keep the files up to date and maintain the security patches, I have updated a few versions of the files. The following preprocessors were modified:
```
1. First Categoriser
2. Second Categoriser
3. Grouping 
```

The `first categoriser and grouping` have been tested on unicorn. However, the `second categoriser` **cannot be tested** on unicorn due to API KEY limitations.

### DIFF
--- a/preprocessors/categoriser/requirements.txt
+++ b/preprocessors/categoriser/requirements.txt
@@ -1,7 +1,7 @@
 flask==1.1.2
 flask_api==2.0.0
-numpy==1.20.1
-pillow==8.2.0
+numpy==1.21.0
+pillow==9.0.0
 jsonschema==3.2.0
 Werkzeug==0.16.0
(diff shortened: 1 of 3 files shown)


jeffbl:
@rohanakut out of curiosity what is the API key limitation?

rohanakut:
I do not have the API key for second categoriser, as the API key is only present on a env file on pegasus.

jeffbl:
OK. I think you have access to pegasus, so should have access to the .env file, but I may be misunderstanding the issue?

rohanakut:
@jeffbl I have tested the second categorsier on unicorn and these changes do not break the pipeline. @jaydeepsingh25 if there aren't any further changes I believe this PR can be merged

jaydeepsingh25 (approved):

PR #297, opened 2022-01-21
Request port 0 when using OSC

### LINKED ISSUE #293: osc.js will reuse the same port (even if it's in use)
By default, [osc.js will bind to port 57121](https://github.com/colinbdclark/osc.js#options-2). This isn't a problem normally, however if two requests reach a handler using osc.js at the same time and it tries to send OSC while another request is in progress, it will attempt to bind to the same port, fail, and error. This may lead to unexpected behavior that is difficult to replicate now, but may become far more problematic as more than one person at a time uses IMAGE.

JRegimbal:
This should be resolved into a free port by the system. This resolves #293. Since testing with a race condition is difficult, I just manually opened a REPL with the osc.js library and opened two ports requesting port 0. Both successfully opened. In other cases (e.g., the default) this would result in an error.

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/handlers/generic-tts-handler/src/server.ts
+++ b/handlers/generic-tts-handler/src/server.ts
@@ -246,7 +246,8 @@ app.post("/handler", async (req, res) => {
         const oscPort = new osc.UDPPort({
             "remoteAddress": "supercollider",
             "remotePort": scPort,
-            "localAddress": "0.0.0.0"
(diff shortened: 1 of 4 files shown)


JRegimbal:
Merging myself since this is a small change to fix a bug.

Issue #294, opened 2022-01-20
Free GPU memory at end of process, even with errors


JRegimbal:
Containers tend to hold onto GPU memory, especially after crashing. This is a two-fold problem:

1. There need to be try-except/catch parts around GPU code to gracefully handle OOM errors.
2. The memory needs to be freed at the end even if an OOM occurs. This would most likely take the form of try-except-finally where we run the code we have now in try, add any graceful error handling in except, and in finally free any memory being taken on the GPU before returning the actual response (whether it's an error or success message).

PR #302, opened 2022-01-26
Added try catch for OOM

rohanakut:
This PR is wrt to issue [294](https://github.com/Shared-Reality-Lab/IMAGE-server/issues/294). The try-catch block have been added to the sections that are generally responsible for the OOM errors. With the current setup, we return a 500 error  every time an OOm is encountered. the GPU memory is also cleared before the exception is thrown. 

The code has been tested on unicorn using the json present in `/home/rakut/test_jsons`

### DIFF
--- a/preprocessors/semanticSeg/segment.py
+++ b/preprocessors/semanticSeg/segment.py
@@ -131,7 +131,7 @@ def run_segmentation(url,
         if 'out of memory' in str(e):
             print("OOM detected")
             torch.cuda.empty_cache()
-            return jsonify("OOM detected"),500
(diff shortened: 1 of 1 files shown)


jeffbl:
To make sure, does this also free the memory as soon as it is done using it, even if there is no OOM? i.e., does it not wait until the next item comes in to do the free?

rohanakut:
@jeffbl yes. Line 253 clears the memory before the response is sent. This would avoid the memory holding problem.

jaydeepsingh25 (approved):

PR #304, opened 2022-01-27
Ocr 114

BenMacnaughton:

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/preprocessors/ocr/ocr.py
+++ b/preprocessors/ocr/ocr.py
@@ -76,8 +76,6 @@ def get_ocr_text():
     name = 'ca.mcgill.a11y.image.preprocessor.ocr'
     request_uuid = content['request_uuid']
     timestamp = int(time.time())
-    logging.error(ocr_result)
-    data = json.dumps(ocr_result)
 
     # try:
     #     validator = jsonschema.Draft7Validator(data_schema, resolver=resolver)
@@ -90,7 +88,9 @@ def get_ocr_text():
         'request_uuid': request_uuid,
         'timestamp': timestamp,
         'name': name,
-        'data': data
+        'data': {
+            'ocr_lines': ocr_result
+        }
     }
 
     try:
@@ -134,7 +134,15 @@ def get_ocr_text(source):
 
     # Check for success
     if read_result.status == OperationStatusCodes.succeeded:
-        return read_result.analyze_result.read_results
+        ocr_results = []
+        for analyzed_result in read_result.analyze_result.read_results:
+            for line in analyzed_result.lines:
+                line_data = {
+                    "text": line.text,
+                    "bounding_box": line.bounding_box
+                }
+                ocr_results.append(line_data)
+        return ocr_results
     else:
         logging.error("OCR text: {}".format(read_result.status))
         return None
--- a/preprocessors/ocr/ocr.py
+++ b/preprocessors/ocr/ocr.py
@@ -68,7 +68,7 @@ def get_ocr_text():
     resolver = jsonschema.RefResolver.from_schema(
             schema, store=schema_store)
     # Get OCR text response
-    ocr_result = get_ocr_text(content['image'])
+    ocr_result = analyze_image(content['image'])
     
     if ocr_result is None:
         return jsonify("Could not retreive Azure results"), 400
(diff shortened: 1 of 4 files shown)


preprocessors/ocr/ocr.py
```diff
@@ -0,0 +1,152 @@
...
+# and our Additional Terms along with this program.
+# If not, see
+# <https://github.com/Shared-Reality-Lab/IMAGE-server/LICENSE>.
+
+
+import json
+import time
+import logging
+import jsonschema
+import requests
```
aidanwilliams09, line 22: ```requests``` module not used

preprocessors/ocr/ocr.py
```diff
@@ -0,0 +1,152 @@
...
+import os
+import io
+import base64
+from flask import Flask, request, jsonify
+from azure.cognitiveservices.vision.computervision import ComputerVisionClient
+from azure.cognitiveservices.vision.computervision.models import OperationStatusCodes
+from azure.cognitiveservices.vision.computervision.models import VisualFeatureTypes
```
aidanwilliams09, line 29: ```VisualFeatureTypes``` module not used

preprocessors/ocr/ocr.py
```diff
@@ -0,0 +1,152 @@
...
+from msrest.authentication import CognitiveServicesCredentials
+
+app = Flask(__name__)
+
+
+@app.route('/preprocessor', methods=['POST', 'GET'])
+def get_ocr_text():
```
aidanwilliams09, line 36: Helper method and api method are called the same thing. Consider disambiguating them

preprocessors/ocr/ocr.py
```diff
@@ -0,0 +1,152 @@
...
+    # Grab the ID from the URL
+    operation_id = read_operation_location.split("/")[-1]
+
+    # Call the "GET" API and wait for it to retrieve the results 
+    # Might get stuck here - need a break condition of some sort
+    while True:
+        read_result = computervision_client.get_read_result(operation_id)
+        if read_result.status not in ['notStarted', 'running']:
+            break
+        time.sleep(1)
```
aidanwilliams09, line 132: Is there a specific reason we need to sleep after calling the api? Also do we know for a fact that every single piece of data will return some status other than these two at some point?

preprocessors/ocr/ocr.py
```diff
@@ -0,0 +1,152 @@
...
+    subscription_key = os.environ["AZURE_API_KEY"]
+    endpoint = "https://image-cv.cognitiveservices.azure.com/"
+    
+    computervision_client = ComputerVisionClient(endpoint, CognitiveServicesCredentials(subscription_key))
+    
+    read_response = computervision_client.read_in_stream(stream,  raw=True)
+    
+    read_operation_location = read_response.headers["Operation-Location"]
+    # Grab the ID from the URL
+    operation_id = read_operation_location.split("/")[-1]
```
aidanwilliams09, line 124: Is it possible to do a regex here? I don't know whether there's a real benefit to it but splitting and getting a specific index is not very error proof.

preprocessors/ocr/ocr.py
```diff
@@ -0,0 +1,152 @@
...
+    return response
+
+def get_ocr_text(source):
+    """
+    Gets OCR text data from Azure API
+    """
+
+    # Convert URI to binary stream
+
+    image_b64 = source.split(",")[1]
```
aidanwilliams09, line 111: See comment on line 124.

preprocessors/ocr/ocr.py
BenMacnaughton: This was the recommendation from the Azure docs. I am not attached to it if you feel like changes are needed

preprocessors/ocr/ocr.py
BenMacnaughton: This was recommended by Azure docs. We do not so I think some sort of time threshold break condition would be suitable

preprocessors/ocr/ocr.py
BenMacnaughton: This is fine. Request schema ensures that this will work

BenMacnaughton:
Tested on several images via curl. Data returns as expected. No end-to-end testing with browser has been done.

BenMacnaughton:
The corresponding schemas PR for the OCR has already been merged

jaydeepsingh25 (approved):

Issue #261, opened 2022-01-10
Schema for Audio-Haptic Semantic Segmentation


sriGanna:
This issue is related to the generation of a new schema that includes the relevant information to render contours on the haply using information from the semantic segmentation preprocessor. 

This would be based on the existing schema for the audio experience for semantic segmentation.

Issue #224, opened 2021-12-08
Clarify Graphics Terminology


JRegimbal:
Currently, the various components of IMAGE use "image" to refer to the IMAGE project, graphics, and photographs. This has already lead to some confusion in conversation and since the project ~will ultimately be~ *has been* open sourced, schemas and public facing parts of containers should be clear in what names are used. Files where "image" is used to refer to graphics or photographs and not the project are in:
* [x] Schemas
  * [x] `request.schema.json`
  * [x] `renderers/simplehaptics.schema.json`
  * [x] `preprocessors/categoriser.json`
* [ ] Preprocessors (tbd)
* [ ] Handlers (tbd)

Cybernide:
I am ALL FOR this issue. Please let me know if I can help.

JRegimbal:
Thanks! If you have any related suggestions please make them! :slightly_smiling_face:

Cybernide:
I've been trying to make these changes on the website for a while now, being specific when it's photographs, and in general terms (when it's not clear whether it's a photo, chart, or map), "graphic". Is the issue that these words need to be changed in the server code?

JRegimbal:
In this specific issue, yes. The different components internally use the terms and include them in the JSON being sent around. User-facing, I think there's a bit less inconsistency, but there may be problems especially in the names of renderings returned to the client (although when we replace those soon it won't be an issue at all).

sriGanna:
Just checking in to see if there is an outcome from this issue? If we were to use this in a schema what is the recommendation in terms of naming? @JRegimbal

JRegimbal:
List of schema fields changed in a breaking way:
* first categorizer: "image" type is now "photograph"
* simple haptics renderer: "image" is now "graphic"
* request: "image" is now "graphic"

JRegimbal:
Changes merged so this issue is closed!

PR #308, opened 2022-01-31
Audiohaptic schema

sriGanna:
This PR is for the schema of an audio-haptic renderer for photos and closes issue #261 . The renderer will send the following data to the extension: 
- the image
- an audio file 
- the segment offset information for the audio file
- the centroid coordinates 
- the contour point coordinates 

This schema has been tested on 3 images: 
- https://en.wikipedia.org/wiki/File:Canyon_no_Lago_de_Furnas.jpg
- [The kitchen example from Image.ally](https://versastyledesign.ca/wp-content/uploads/2020/03/Montreal-Kitchen-Design-04_1.jpg)
- [The outdoor scene from Image.ally](https://versastyledesign.ca/wp-content/uploads/2020/03/Montreal-Kitchen-Design-04_1.jpg)


renderers/photoaudiohaptics.schema.json
```diff
@@ -0,0 +1,91 @@
...
+            "items": {
+                "type": "number",
+                "minimum": 0,
+                "maximum": 1
+            }
+        },
+        "contourPoints": {
+            "type": "array",
+            "items": {
+                "allOf": [
```
JRegimbal, line 29: Shouldn't be `allOf`s with one item.

renderers/photoaudiohaptics.schema.json
```diff
@@ -0,0 +1,91 @@
...
+    },
+    "properties": {
+        "image": {
+            "type": "string"
+        },
+        "info": {
+            "type": "object",
+            "properties": {
+                "items": {
+                    "allOf": [
```
JRegimbal, line 45: Single element `allOf`

renderers/photoaudiohaptics.schema.json
```diff
@@ -0,0 +1,91 @@
...
+                                            {
+                                                "$ref": "./definitions.json#/definitions/segmentOffsetSeconds"
+                                            },
+                                            {
+                                                "type": "object",
+                                                "properties": {
+                                                    "name": {
+                                                        "description": "A short name assigned to the segment.",
+                                                        "type": "string",
+                                                        "maxLength": 300
```
JRegimbal, line 65: Is 300 characters short? What's the purpose of this? If it's just an identifier then this seems excessive to me...

renderers/photoaudiohaptics.schema.json
```diff
@@ -0,0 +1,91 @@
...
+            "properties": {
+                "items": {
+                    "allOf": [
+                        {
+                            "type": "object",
+                            "properties": {
+                                "audioFile": {
+                                    "$ref": "./definitions.json#/definitions/audioFile"
+                                },
+                                "entityInfo": {
```
JRegimbal, line 52: What's an entity in this context?

renderers/photoaudiohaptics.schema.json
```diff
@@ -0,0 +1,91 @@
...
+                "allOf": [
+                    {
+                        "$ref": "#/definitions/coordinates"
+                    }
+                ]
+            }
+        }
+    },
+    "properties": {
+        "image": {
```
JRegimbal, line 38: I don't think you need this? I'm loathe to include the graphic (deprecating "image" use see #224), but if you do really need it then there needs to be more clarification as to format than "string".

JRegimbal (changes requested):
First pass on it. There's a very sparse amount of comments/description here which makes it hard to understand. Those would certainly help. Also, you've created a new schemas submodule *in the schemas*. Please delete this.

renderers/photoaudiohaptics.schema.json
```diff
@@ -0,0 +1,91 @@
...
+            "type": "array",
+            "items": {
+                "type": "number",
+                "minimum": 0,
+                "maximum": 1
+            },
+            "minItems": 2,
+            "maxItems": 2
+        },
+        "coordinates": {
```
JRegimbal, line 18: Not required since this is very single use, but previously we've encoded each coordinate pair as a tuple and then had longer lists of coordinate pairs be arrays of those tuples. This helps avoid problems with, for example, an odd number of coordinates being submitted by accident and having that pass validation.

renderers/photoaudiohaptics.schema.json
sriGanna: to clarify, are you suggesting we shorten it or get rid of the constraint?

renderers/photoaudiohaptics.schema.json
sriGanna: Tagging @Cybernide  for input as well but based on some previous convos having the image in the haptic rendering would be useful for user testing.

renderers/photoaudiohaptics.schema.json
sriGanna: given that this could potentially require significant reworking would it be possible to address the other issues before we dedicate time to this?

sriGanna:
@JRegimbal we updated the based on your comments. we need further clarification on the rename of the "image" identifier

renderers/photoaudiohaptics.schema.json
JRegimbal: Sure, although note that that's the output format used by semantic segmentation currently so you would have the additional task of processing that.

renderers/photoaudiohaptics.schema.json
JRegimbal: Not saying don't include the image, just that there should really be a pipeline to accessing it in the extension...it doesn't make sense to me for the extension to send a blob and then the server to return that same blob unmodified.

renderers/photoaudiohaptics.schema.json
JRegimbal: Well what are examples of `name`? I think a length constraint is good just 300 may be excessive.

JRegimbal:
We're trying to move away from using "image" to just refer to any old graphical media since it can be confused with IMAGE the project. Since it's a file, I'd recommend using "graphic".

renderers/photoaudiohaptics.schema.json
```diff
@@ -0,0 +1,86 @@
...
+            "properties": {
+                "items": {
+                    "type": "object",
+                    "properties": {
+                        "audioFile": {
+                            "$ref": "./definitions.json#/definitions/audioFile"
+                        },
+                        "entityInfo": {
+                            "type": "array",
+                            "description":  "An entity contains the relevant audio-haptic information to locate each segment or object.",
```
JRegimbal, line 53: To clarify, an entity is data about an object or segment? How is this ordered?

renderers/photoaudiohaptics.schema.json
Cybernide: Ah yes, Sri, we should be referring to them as _graphics_ from here on in to avoid confusion. In fact, do read #224, which which involves refactoring all references from "images" to "graphics".

But I do agree that it would be good to have a copy of the graphic in the rendering "area?" it would help partially sighted users. Can we make a work item to make this pipeline?

rayanisran: Makes sense. I think we can probably pass the blob directly into the pop-up menu from the browser.

renderers/photoaudiohaptics.schema.json
rayanisran: Segments, followed by objects. Could be clarified in the description.

renderers/photoaudiohaptics.schema.json
```diff
@@ -34,10 +34,10 @@
         }
     },
     "properties": {
-        "image": {
+        "graphic": {
             "type": "string",
-            "description": "A base64 encoding of the image as a string.",
-            "pattern": "data:image\/[^;]+;base64[^\"]+"
+            "description": "A base64 encoding of the graphic as a string.",
+            "pattern": "data:graphic\/[^;]+;base64[^\"]+"
```
JRegimbal, line 40: This is a [media type](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types).
```suggestion
            "pattern": "data:image\/[^;]+;base64[^\"]+"
```

JRegimbal (approved):
Mostly seems fine to me. Is it working with the handler?

sriGanna:
Yup, we've tested the latest changes to the schema with the handler

PR #315, opened 2022-02-02
remove restrictive pattern on api_request format as it is causing iss…

BenMacnaughton:
…ues with validation and poses no threat to cause any downstream problems

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/preprocessors/autour.schema.json
+++ b/preprocessors/autour.schema.json
@@ -22,7 +22,8 @@
         "api_request": {
             "type": "string",
             "title": "API Request",
-            "description": "The API request to be made to the map"
+            "description": "The API request to be made to the map",
+            "pattern": "^https?://(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{2,}\\.[a-z]{2,}\\b(\\S*)$"
         },
         "places": {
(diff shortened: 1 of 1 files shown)


JRegimbal (commented):
Don't really know what the issue is! We have updated the [URL definition used elsewhere](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/47be5c8b5c0a563513ea021c0844ff140ada0987/definitions.json#L31) because they ended up being too restrictive before. Would that work for your case?

BenMacnaughton:
Unfortunately, it does not since the extra arguments in the API call cause it to fail. I have just tested it and think that it might be best to leave the field unrestricted since there are a number of arguments that we may add or remove.

JRegimbal:
Can you give an example of a possible API call?

BenMacnaughton:
After looking around a bit it turned out to be a whitespace issue in the autour preprocessor that I will update. Looks like the pattern you referenced should be good for this.

JRegimbal:
Looks like I gave you the same pattern we started with without noticing...closing since there's now no difference between this branch and schemas.

PR #317, opened 2022-02-03
Removing whitespace from Autour API call

BenMacnaughton:

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/preprocessors/autour/autour.py
+++ b/preprocessors/autour/autour.py
@@ -86,6 +86,8 @@ def get_map_data():
             font=9&\
             pad=0"
 
+    api_request = ''.join(api_request.split())
+
     response = requests.get(api_request).json()
     results = response['results']
 


BenMacnaughton:
Just adding the line `api_request = ''.join(api_request.split())` to rid API call of whitespace which was causing trouble

BenMacnaughton:
Has been tested using CURL on Unicorn. Since this was an extremely minor change I believe this to be sufficent.

preprocessors/autour/autour.py
```diff
@@ -86,6 +86,8 @@ def get_map_data():
             font=9&\
             pad=0"
 
+    api_request = ''.join(api_request.split())
```
JRegimbal, line 89: 1) This should probably be assigned to @jaydeepsingh25 
2) Does this alter the content of the request? If the issue is whitespace, it would probably be sufficient to encode each space as `%20`.

BenMacnaughton: Ok sounds good I will change the assignee. No content changes for the request

jaydeepsingh25 (approved):

Issue #314, opened 2022-02-01
Add functionality to temporarily store user data for opt-in support


JRegimbal:
In order to provide meaningful support to people who submit feedback/bugs, we need a way to access request and response data when given the user's consent. It would also be more convenient to the user to give consent for a particular request after something noteworthy occurred rather than to give blanket consent on all requests after a point in time. For this to work, all requests and responses would need to be saved for a short amount of time (~1 hour max.) to provide a window to the user before this data is permanently deleted.

This data would need to be associated to a particular request_uuid and a periodic job would need to be created to purge old files that do not contain a marker indicating the user has consented to their continued storage and use. This issue is solely concerned with creating the functionality, off by default, to store the data in a suitable format so it can be used later.

jeffbl:
Putting in Open Beta for now, pending permissions and approvals.

JRegimbal:
Currently running on Unicorn with logging disabled.

PR #319, opened 2022-02-03
Temporary logging

JRegimbal:
Backend functionality for #314. This adds the following changes:

- a new volume for storing user requests and responses
- user requests and responses are automatically saved for 1 hour when the `STORE_IMAGE_DATA` environment variable is set to `on` or `ON`
- an endpoint is created at `/authenticate/:uuid/:check` to allow users to mark a temporary saved request/response pair for long-term storage by sending a GET request to the endpoint with the request UUID and a SHA1 sum of the request object using [object-hash](https://www.npmjs.com/package/object-hash)
- runs a cron job every 10 minutes to delete any logs at least an hour old that do not have a file marking that a user successfully authenticated.

This was tested locally by sending different requests and messages. The requests save and clear as expected when unmarked. When marked, they do not get cleared. The authentication endpoint does not mark when the checksum does not match and will only return a 401 Unauthorized error so long as the UUID is formatted correctly. 200 OK is returned upon success. Note that this function can be run multiple times without side effects.

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,11 +8,14 @@ services:
         image: "orchestrator:latest"
         env_file:
             - ./config/express-common.env
+        environment:
+            - STORE_IMAGE_DATA=ON
         group_add:
             # This should be set in an environment variable or .env
             - ${DOCKER_GID}
         volumes:
-            - /var/run/docker.sock:/var/run/docker.sock
+            - /var/run/docker.sock:/var/run/docker.sock:ro
+            - user-logs:/var/log/IMAGE
     supercollider:
         build:
             context: ./services/supercollider-images/supercollider-service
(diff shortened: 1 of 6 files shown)


JRegimbal:
For anyone curious, the cron job function split across multiple lines for readability.
```sh
for f in $(find /var/log/IMAGE/ -mindepth 1 -maxdepth 1 -type d); do 
  if [ ! -f $f/auth ]; then
    if [ $(($(date +%s) - $(date +%s -r $f))) -gt 3600 ]; then
      rm -r $f;
      echo "Removed $f" >> /proc/1/fd/1;
     fi;
  fi;
done;
```

jaydeepsingh25 (approved):

PR #338, opened 2022-02-09
Build smaller images

JRegimbal:
This is an effort to reduce the size of our images for download. Some will just always be large (the base YOLO image is several GB on its own), but some don't need to be. This is PR focuses on removing unused dependencies and using smaller base images.

All images in commits have been tested locally. If you're tagged as a reviewer, it's because the image belongs to you and you should check that this won't have odd side effects.

---

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/build.yml
+++ b/build.yml
@@ -111,6 +111,7 @@ services:
         build:
             context: .
             dockerfile: ./handlers/photo-audio-haptics-handler/Dockerfile
+        image: "photo-audio-haptics-handler:latest"
     # Supercollider on Fedora 34
     supercollider-base:
         build:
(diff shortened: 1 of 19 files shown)


BenMacnaughton (changes requested):
I believe that the pybase64 dependency is needed to decode the image data for the Azure API. Unless I am misunderstanding I don't think the base64 library is native

JRegimbal:
@BenMacnaughton base64 is native to Python (https://docs.python.org/3/library/base64.html)

BenMacnaughton:
@JRegimbal apologies for the misunderstanding the rest looks good to me

BenMacnaughton (approved):

sriGanna (approved):

Issue #345, opened 2022-02-15
Semseg has value greater than 1.0 which fails schema check


jeffbl:
While getting rendering for a graphic on the tutorial page, got this from semantic segmentation:

```
atp-object-detection           | 2022-02-15T01:48:15.102660765Z [2022-02-15 01:48:15 +0000] [31] [DEBUG] POST /preprocessor
atp-object-detection           | 2022-02-15T01:48:15.218480444Z YOLOv5 🚀 v6.0-0-g956be8e torch 1.9.1+cu102 CPU                                                                                                    
atp-object-detection           | 2022-02-15T01:48:15.218514037Z                                                                                                                                                    
atp-object-detection           | 2022-02-15T01:48:15.381605237Z Fusing layers...                                                                                                                                   
atp-object-detection           | 2022-02-15T01:48:16.770197863Z Model Summary: 476 layers, 87730285 parameters, 0 gradients
atp-semantic-segmentation      | 2022-02-15T01:48:17.107285883Z [2022-02-15 01:48:17 +0000] [9] [DEBUG] POST /preprocessor               
atp-semantic-segmentation      | 2022-02-15T01:48:18.875655124Z ERROR:root:1.0007072135785007 is greater than the maximum of 1                         
atp-semantic-segmentation      | 2022-02-15T01:48:18.875693646Z                                                                                                                                                    
atp-semantic-segmentation      | 2022-02-15T01:48:18.875699356Z Failed validating 'maximum' in schema['properties']['segments']['items']['properties']['contours']['items']['properties']['coordinates']['items']['
items']:                                                                                                                                                                                                           
atp-semantic-segmentation      | 2022-02-15T01:48:18.875705077Z     {'maximum': 1, 'minimum': 0, 'type': 'number'}                                                                                                 
atp-semantic-segmentation      | 2022-02-15T01:48:18.875709165Z                                                                                                                                                    
atp-semantic-segmentation      | 2022-02-15T01:48:18.875712922Z On instance['segments'][0]['contours'][8]['coordinates'][1504][1]:
atp-semantic-segmentation      | 2022-02-15T01:48:18.875717871Z     1.0007072135785007                                                                                                                             
atp-orchestrator               | 2022-02-15T01:48:18.876959801Z Error occured on fetch from semantic-segmentation         
atp-orchestrator               | 2022-02-15T01:48:18.876999926Z Invalid Preprocessor JSON format                                                                                                                   
[...]

Issue #348, opened 2022-02-15
semseg scaling seems to not respect aspect ratio


jeffbl:
Looking at the semseg code, there are some possible weirdnesses with the scaling and dimensions:

https://github.com/Shared-Reality-Lab/IMAGE-server/blob/e5fc06a84db0c85ec1d4c55f9990c8af497f1faa/preprocessors/semanticSeg/segment.py#L140

If I'm reading that correctly, it seems that it is scaling both axes to 1500 pixels if either axis exceeds 1500 pixels. This is odd, since it will almost certainly distort the aspect ratio of anything except an already square graphic. This could potentially distort features by quite a bit, impacting the ML algorithm? It will also make the overall graphic larger in some cases, e.g., if the original is 1502x750, it'll make it 1500x1500, which will be like twice the number of pixels as the original? If this is the case, there are better ways to do the scaling, with maintaining the aspect ratio, e.g.:

https://www.tutorialkart.com/opencv/python/opencv-python-resize-image/

Further, I don't see in that function where the width/height are reset after resizing, such that they would be the new image size after scaling? However, those variables are used later in the function:
https://github.com/Shared-Reality-Lab/IMAGE-server/blob/e5fc06a84db0c85ec1d4c55f9990c8af497f1faa/preprocessors/semanticSeg/segment.py#L163

I am not sure if these are relevant to semseg failing due to schema failure (#345), but I'm logging separately.

jeffbl:
The width/height is now reset from #352. So this issue is now only for the scaling not preserving aspect ratio. Moving forward to Open Beta since we don't know how large an impact this has on semseg results, but my guess is it distorts them.

PR #352, opened 2022-02-16
Changed height and width to ensure proper normalisation

rohanakut:
This PR is wrt to #345 . The code has been tested on unicorn against the image mentioned in #345. the normalisation values are between 0 and 1 and hence a valid response is generated for the code


jeffbl (approved):
This is OK only as a temporary fix until #348 is fixed, which should be done as soon as possible.

jaydeepsingh25 (approved):

jeffbl:
@rohanakut:  @jaydeepsingh25 and I both missed that you had reverted schemas to an earlier version in this PR. This caused mismatched schemas. Please be careful to not check in reversions.

PR #357, opened 2022-02-17
Preserved aspect ratio

rohanakut:
This PR is wrt to #348. The code has been tested on unicorn with the 3rd image on tutorial page and the 2nd image in the home page.

### DIFF
--- a/preprocessors/semanticSeg/segment.py
+++ b/preprocessors/semanticSeg/segment.py
@@ -137,45 +137,11 @@ def run_segmentation(url,
     pil_image = cv2.imdecode(image, cv2.IMREAD_COLOR)
     height, width, channels = pil_image.shape
     scale_size = np.float(1500.0 / np.float(max(height,width)))
-    if(height>1500 or width >1500):
+    if(scale_size<1.0):
         height = np.int(height * scale_size)
         width = np.int(width * scale_size)
         pil_image = cv2.resize(pil_image, (width, height),
                             interpolation=cv2.INTER_AREA)
-    # if(scale_size < 1.0):    # graphic too big! it needs to be scaled down
-    #     height = np.int(height * scale_size)
-    #     width = np.int(width * scale_size)
-    #     pil_image = cv2.resize(pil_image, (width, height),
-    #                         interpolation=cv2.INTER_AREA)
-    #safer and probably faster to just get new dimensions
-    #  from the actual new image, e.g., avoids rounding differences
-    # height, width, channels = pil_image.shape
-    # if(height > 1500 or width > 1500):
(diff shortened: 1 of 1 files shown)


jeffbl (changes requested):
I'm looking at the code change in segment.py so will make another review comment for that, but already I see a huge number of scehma changes again. This was a big problem that cost a bunch of time in https://github.com/Shared-Reality-Lab/IMAGE-server/pull/352. Please ping @JRegimbal if you need help on this.

preprocessors/semanticSeg/segment.py
```diff
@@ -137,10 +137,31 @@ def run_segmentation(url,
...
+                height = np.int(height * scale_size)
+                width = np.int(width * scale_size)
+                pil_image = cv2.resize(pil_image, (width, height),
+                                       interpolation=cv2.INTER_AREA)
+            else:
+                scale_size = np.float(1500.0 / np.float(width))
+                height = np.int(height * scale_size)
+                width = np.int(width * scale_size)
+                pil_image = cv2.resize(pil_image, (width, height),
+                                       interpolation=cv2.INTER_AREA)
```
jeffbl, line 164: This is a large amount of code for the task, with a constant (1500) hardcoded in many different places, with multiple nested if/elses that are easy to get wrong and hard to read due to repetitive code. I think something like the following (uncompiled! not tested!) would be both clearer and much shorter, so less prone to bugs if it needs to be changed in the future:

```
  scale_size = np.float(1500 / np.float(max(height,width))
     if(scale_size < 1.0):    # graphic too big! it needs to be scaled down
        height = np.int(height * scale_size)
        width = np.int(width * scale_size)
        pil_image = cv2.resize(pil_image, (width, height),
                            interpolation=cv2.INTER_AREA)
       #safer and probably faster to just get new dimensions
       #  from the actual new image, e.g., avoids rounding differences
       height, width, channels = pil_image.shape
```

jeffbl (changes requested):
Removed since was a draft of below code recommendation

jeffbl (commented):
code looks correct to me from inspection.

rohanakut:
@jeffbl the code has been tested on 3 tutorial images and the resizing is working as expected.

jeffbl (changes requested):
Why are there 16 schema files being updated again?
Code looks OK to me assuming you have verified that cv2.resize is guaranteed to return an image of the requested size, even in error conditions. (otherwise height/width should be read back from the returned image to make sure they are in sync)

jeffbl (approved):

jaydeepsingh25 (approved):

PR #368, opened 2022-02-24
Deprecate confidence in renderings

### LINKED ISSUE #203: Determine future of "confidence" in the architecture
Currently, confidence is meaningless. The values from preprocessors, if any, are garbage. The handlers can't actually use confidence for preprocessors to do anything and the handlers' own advertized confidence suffers from "garbage in garbage out".

This work item focuses on determining what should be done with "confidence", whether we redefine it, change the role, or remove it entirely from the schemas.

JRegimbal:
So far, we haven't been able to meaningfully create a confidence for an entire rendering. In case we do it remains in the schema (plus for backwards compatibility), but otherwise we mark it as deprecated and do not require it. Resolves #203. Confidence may continue as a per-preprocessor report.

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/definitions.json
+++ b/definitions.json
@@ -56,7 +56,7 @@
             "properties": {
                 "type_id": { "$ref": "#/definitions/renderer" },
                 "confidence": {
-                    "description": "Estimated percent confidence in the correctness of the rendering.",
+                    "description": "DEPRECATED: Estimated percent confidence in the correctness of the rendering.",
                     "type": "number",
                     "minimum": 0,
                     "maximum": 100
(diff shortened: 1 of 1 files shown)


JRegimbal:
Assigning to Jeff to check whether this seems like a reasonable solution for now.

definitions.json
```diff
@@ -56,7 +56,7 @@
             "properties": {
                 "type_id": { "$ref": "#/definitions/renderer" },
                 "confidence": {
-                    "description": "Estimated percent confidence in the correctness of the rendering.",
+                    "description": "DEPRECATED: Estimated percent confidence in the correctness of the rendering.",
```
jeffbl, line 59: We don't currently do anything with this, so agree it is important that it be flagged since it could otherwise be confusing, but I suspect we may return to this at some point, so I view it more as a placeholder than something that we did at one point and are removing forever. Suggest:

"DEPRECATED (but reserved for future use): Estimated..."

jeffbl, line 59:
```suggestion
                    "description": "DEPRECATED (but reserved for future use): Estimated percent confidence in the correctness of the rendering.",
```

PR #373, opened 2022-03-01
Ah schema edits

sriGanna:
This PR reflects changes to the Photo-audio-haptics-handler in response to changes to the semantic segmentation. This schema was tested on unicorn against most of the pictures of the image Demo Page as well as the following image:
https://en.wikipedia.org/wiki/File:Canyon_no_Lago_de_Furnas.jpg

## Required Information

## Pull Request Checklist

- [ ] The Docker image builds


renderers/photoaudiohaptics.schema.json
```diff
@@ -25,12 +25,16 @@
                 "maximum": 1
             }
         },
-        "contourPoints": {
+        "contours": {
```
JRegimbal, line 28: To clarify, this is all the points with a label and not broken down by individual contours, right?

renderers/photoaudiohaptics.schema.json
```diff
@@ -25,12 +25,16 @@
...
             "type": "array",
             "description": "An array containing an array of normalized coordinates representing a point on the segment or point on the bounding box.",
             "items": {
                 "$ref": "#/definitions/coordinates"
             }
+        },
+        
+        "entityType":{
```
JRegimbal, line 36: Could you add more documentation for this? A description is a must, and if this only takes a few values then it should probably just be enumerated.

JRegimbal (changes requested):

renderers/photoaudiohaptics.schema.json
sriGanna: Sorry, I forgot to add the description! I'll also add a max length for the string

renderers/photoaudiohaptics.schema.json
sriGanna: yes

renderers/photoaudiohaptics.schema.json
```diff
@@ -25,12 +25,18 @@
...
+            "type": "string",
+            "description": "This describes whether the entity is a segment, object or static text",
```
JRegimbal, line 38: Based on the PR with the handler code, it looks like there are only a few values for this. Could you change the type to an enumeration like below?
https://github.com/Shared-Reality-Lab/IMAGE-server/blob/ff3453c2d4974575f0f6b6681232184462c5b0e5/preprocessors/content-categoriser.schema.json#L8

JRegimbal (changes requested):
One last change to this with just enumerating the entity types.

renderers/photoaudiohaptics.schema.json
sriGanna: yes, that would make sense!

JRegimbal (approved):

PR #376, opened 2022-03-02
Add support for HighCharts in IMAGE Request

jaydeepsingh25:

## Required Information

## Pull Request Checklist

- [ ] The Docker image builds


request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+                    "properties": {
+                        "description":{
+                            "description": "Description text about y-axis",
+                            "type": "string"
+                        }
+                    }
+                },
+                "noOfDataPoints": {
+                    "type": "number",
+                    "description": "Maximum number of data points present in all the series in the graph"
```
jeffbl, line 60:
```suggestion
                    "description": "Number of data points in the longest series in the graph"
```

request.schema.json
```diff
@@ -12,6 +12,105 @@
             "type": "string",
             "pattern": "^data:image/[^;]+;base64,[a-zA-Z0-9+/=]+$"
         },
+        "highChartsData": {
```
jeffbl, line 15: If this format would potentially be generic enough to be used for data from other charts, suggest naming it more generically, e.g., just "chartData"? @JRegimbal - thoughts?

request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+            "description":"Data extracted from the highchart",
```
jeffbl, line 16:
```suggestion
            "description":"Data extracted from the chart",
```

request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+                            "description": "Description text about x-axis",
+                            "type": "string"
+                        }
+                    }
+                },
+                "yAxis":{
+                    "description":"Information about y-axis of the graph",
+                    "type": "object",
+                    "properties": {
+                        "description":{
```
jeffbl, line 52: It seems odd to me that the x-axis gets "labels" but not the y-axis?

request.schema.json
JRegimbal: This seems to just be the plain data taken from highcharts. I'd argue we shouldn't give it a generic name since it would conflict with any truly generic input that's accepted later.

request.schema.json
JRegimbal: I'd advise against this if it is in fact just the highchart format.

request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+                    "type":"string",
+                    "description": "Title of the graph"
+                },
+                "subtitle":{
+                    "type":"string",
+                    "description":"Sub title of the graph"
+                },
+                "type":{
+                    "type": "string",
+                    "description": "Type of graph (line/pie/bar)"
```
JRegimbal, line 29: If it's only these values, type should be an enumeration.

request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+                },
+                "xAxis":{
+                    "description":"Information about x-axis of the graph",
+                    "type": "object",
+                    "properties": {
+                        "labels":{
+                            "description": "Labels present on the x-axis",
```
JRegimbal, line 36: So these are tick labels and not an overall label for the axis, right?

request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+                            "type" : "array",
+                            "items": {
+                                "type":"string"
+                            }
+                        },
+                        "description":{
+                            "description": "Description text about x-axis",
```
JRegimbal, line 43: I'm not sure what this is supposed to be.

request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+                            "type": "string"
+                        }
+                    }
+                },
+                "noOfDataPoints": {
+                    "type": "number",
+                    "description": "Maximum number of data points present in all the series in the graph"
+                },
+                "data":{
+                    "description": "Collection of all the series in the graph",
```
JRegimbal, line 63: This only includes the series property. Is this a simplification of the original format?

request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+                            "type":"array",
+                            "items":{
+                                "type" : "object",
+                                "properties": {
+                                        "name":{
+                                            "description":"Information represented by the series",
+                                            "type": "string"
+                                        },
+                                        "type":{
+                                            "description":"Type of graph represented by series (line/bar/pie)",
```
JRegimbal, line 77: Same enumeration point as before, and can this differ from the type key in the root of this structure?

request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+                                        },
+                                        "data":{
+                                            "description":"Data points in the series",
+                                            "type":"array",
+                                            "items": {
+                                                "type":"object",
+                                                "properties": {
+                                                    "x":{
+                                                        "description":"x-axis value represented by the point",
+                                                        "type":["number","null"]
```
JRegimbal, line 88: Why null? Applies to all the other fields.

request.schema.json
```diff
@@ -12,6 +12,105 @@
...
+                                                    },
+                                                    "y":{
+                                                        "description":"y-axis value represented by the point",
+                                                        "type":["number","null"]
+                                                    },
+                                                    "category":{
+                                                        "description":"Category represented the x-axis value",
```
JRegimbal, line 95: What "categories"?

JRegimbal (changes requested):
Lots of this is very unclear to me. Honestly, if this is just HighCharts input data in its raw form, I don't see the point in attempting to create a well defined schema for it right now, especially if we are going to make simplifications in the format. It just seems like we're asking for compatibility problems. Is there an actual pressing need for anything more than below?
```
"highChartsData": {
    "type": "object",
    "description": "Chart data in the HighCharts format (some url)"
}
```

request.schema.json
jeffbl: OK, I am fine with locking this to highcharts, and revisiting after CSUN when we more deeply address a generic format for charts that could be used for multiple chart sources.

request.schema.json
```diff
@@ -116,6 +120,16 @@
                 "capabilities",
                 "renderers"
             ]
+        },
+        {
+            "required": [
+                "request_uuid",
+                "language",
+                "timestamp",
+                "highChartsData",
```
JRegimbal, line 129: I'd include the context field here as well. It may be useful for stuff like a surrounding caption/title. But not really blocking.

JRegimbal (approved):
Seems good to me. Since @jeffbl has already weighed in I'll hold off on merging so he can have a look.

jeffbl (approved):

PR #384, opened 2022-03-03
Azure od

gp1702:

## Required Information

## Pull Request Checklist

- [ ] The Docker image builds

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -88,6 +88,11 @@ services:
                     devices:
                     - driver: nvidia
                       capabilities: ["gpu", "utility", "compute"]
+    object-detection-azure:
+        image: ghcr.io/shared-reality-lab/image-preprocessor-object-detection-azure:unstable
+        labels:
+            ca.mcgill.a11y.image.preprocessor: 6
+            ca.mcgill.a11y.image.port: 5000
     # Handlers
     autour-handler:
         image: ghcr.io/shared-reality-lab/image-handler-autour:unstable
--- a/preprocessors/object-detection-azure/objdetect.py
+++ b/preprocessors/object-detection-azure/objdetect.py
@@ -1,0 +1,190 @@
+# Copyright (c) 2021 IMAGE Project, Shared Reality Lab, McGill University
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# and our Additional Terms along with this program.
+# If not, see
+# <https://github.com/Shared-Reality-Lab/IMAGE-server/LICENSE>.
+
+import requests  # pip3 install requests
+from re import search
+import operator
+
+# import numpy as np
+import json
+import time
+import jsonschema
+import logging
+import base64
+import os
+from flask import Flask, request, jsonify
+
+app = Flask(__name__)
+
+
+def process_results(result):
+    send = []
+    image_height = result["metadata"]["height"]
+    image_width = result["metadata"]["width"]
+    for item in range(len(result['objects'])):
+        dims = result["objects"][item]['rectangle']
+        x_left = dims['x'] / image_width
+        x_right = (dims['x'] + dims['w']) / image_width
+        y_left = dims['y'] / image_height
+        y_right = (dims['y'] + dims['h']) / image_height
+        centre = [abs((x_left + x_right) / 2),
+                  abs((y_left + y_right) / 2)]
+        area = abs((dims['w'] * dims['h']) / (image_height * image_width))
+        class_prob = float(result['objects'][item]['confidence'])
+        normalised_dims = [x_left, y_left, x_right, y_right]
+        dictionary = {"ID": item,
+                      "type": result['objects'][item]['object'],
+                      "dimensions": normalised_dims,
+                      "confidence": class_prob,
+                      "centroid": centre,
+                      "area": area
+                      }
+        send.append(dictionary)
+    return send
+
+
+def process_image(image):
+
+    region = "canadacentral"  # For example, "westus"
+    api_key = os.environ["AZURE_API_KEY"]
+
+    # Set request headers
+    headers = dict()
+    headers['Ocp-Apim-Subscription-Key'] = api_key
+    headers['Content-Type'] = 'application/octet-stream'
+
+    # Set request querystring parameters
+    """Only query categories"""
+    params = {'visualFeatures': 'Objects'}
+
+    # Make request and process response
+    response = requests.request(
+        'post',
+        "https://{}.api.cognitive.microsoft.com/vision/v1.0/analyze".format(region),
+        data=image,
+        headers=headers,
+        params=params
+    )
+    if response.status_code == 200 or response.status_code == 201:
+
+        if 'content-length' in response.headers and \
+                int(response.headers['content-length']) == 0:
+            return "", 204
+        elif 'content-type' in response.headers and \
+                isinstance(response.headers['content-type'], str):
+            if 'application/json' in response.headers['content-type'].lower():
+                if response.content:
+                    result = response.json()
+                    label = process_results(result)
+                else:
+                    return "", 204
+            else:
+                return "", 204
+
+    else:
+        return "", 204
+
+    return label
+
+
+
+@app.route("/preprocessor", methods=['POST', ])
+def categorise():
+    logging.debug("Received request")
+    # load the schema
+    with open('./schemas/preprocessors/object-detection.schema.json') \
+            as jsonfile:
+        data_schema = json.load(jsonfile)
+    with open('./schemas/preprocessor-response.schema.json') \
+            as jsonfile:
+        schema = json.load(jsonfile)
+    with open('./schemas/definitions.json') as jsonfile:
+        definitionSchema = json.load(jsonfile)
+    with open('./schemas/request.schema.json') as jsonfile:
+        first_schema = json.load(jsonfile)
+    schema_store = {
+        schema['$id']: schema,
+        definitionSchema['$id']: definitionSchema
+    }
+    resolver = jsonschema.RefResolver.from_schema(
+        schema, store=schema_store)
+
+    content = request.get_json()
+    try:
+        validator = jsonschema.Draft7Validator(first_schema, resolver=resolver)
+        validator.validate(content)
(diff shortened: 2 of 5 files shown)


preprocessors/object-detection-azure/objdetect.py
```diff
@@ -0,0 +1,190 @@
...
+    content = request.get_json()
+    try:
+        validator = jsonschema.Draft7Validator(first_schema, resolver=resolver)
+        validator.validate(content)
+    except jsonschema.exceptions.ValidationError as e:
+        logging.error(e)
+        return jsonify("Invalid Preprocessor JSON format"), 400
+    request_uuid = content["request_uuid"]
+    timestamp = time.time()
+    name = "ca.mcgill.a11y.image.preprocessor.objectDetectionAzure"
```
JRegimbal, line 134: Wrong name. Should be objectDetection.

preprocessors/object-detection-azure/objdetect.py
```diff
@@ -0,0 +1,190 @@
...
+                      "area": area
+                      }
+        send.append(dictionary)
+    return send
+
+
+def process_image(image):
+
+    region = "canadacentral"  # For example, "westus"
+    api_key = os.environ["AZURE_API_KEY"]
```
JRegimbal, line 62: Crashes if API key is missing. Should fail gracefully.

preprocessors/object-detection-azure/objdetect.py
```diff
@@ -0,0 +1,190 @@
...
+            classifier_1_label = \
+                classifier_1_output["category"]
+            if classifier_1_label == "photograph":
+                source = content["graphic"]
+                image_b64 = source.split(",")[1]
+                binary = base64.b64decode(image_b64)
+                pred = process_image(image=binary)
+                type = {"objects": pred}
+            else:
+                """If the first classifier does not detect a photograph
```
JRegimbal, line 153: Old names (first, second classifiers)

preprocessors/object-detection-azure/objdetect.py
```diff
@@ -0,0 +1,190 @@
...
+
+    return label
+
+
+
+@app.route("/preprocessor", methods=['POST', ])
+def categorise():
+    logging.debug("Received request")
+    # load the schema
+    with open('./schemas/preprocessors/object-detection.schema.json') \
```
JRegimbal, line 108: Why are schemas being loaded on each request? These files won't change and the resolver can just be reused.

preprocessors/object-detection-azure/objdetect.py
```diff
@@ -0,0 +1,190 @@
...
+        else:
+            """We are providing the user the ability to process a photograph
+            even when the first classifier is absent, however it is
+            recommended that the second classifier be used in conjunction
+            with the first classifier."""
+            source = content["graphic"]
+            image_b64 = source.split(",")[1]
+            binary = base64.b64decode(image_b64)
+            pred = process_image(image=binary)
+            type = {"objects": pred}
```
JRegimbal, line 165: Lots of labels have varying levels of specificity ("plant" vs "house plant" being returned in the same image). Do we get any more information back from azure?

preprocessors/object-detection-azure/objdetect.py
```diff
@@ -0,0 +1,190 @@
...
+    if "graphic" not in content.keys():
+        return "", 204
+    else:
+        if classifier_1 in preprocess_output:
+            classifier_1_output = \
+                preprocess_output[classifier_1]
+            classifier_1_label = \
+                classifier_1_output["category"]
+            if classifier_1_label == "photograph":
+                source = content["graphic"]
```
JRegimbal, line 147: This code section should really be in a separate function since it's reused below.

preprocessors/object-detection-azure/objdetect.py
```diff
@@ -0,0 +1,190 @@
...
+            if 'application/json' in response.headers['content-type'].lower():
+                if response.content:
+                    result = response.json()
+                    label = process_results(result)
+                else:
+                    return "", 204
+            else:
+                return "", 204
+
+    else:
```
JRegimbal, line 97: If we got a not ok response from azure, this really needs to be logged at a warning/error level. As of now, it could fail silently.

JRegimbal: Note that this applies for all of the `return "", 204` lines in this part of the file.

preprocessors/object-detection-azure/objdetect.py
JRegimbal: Also some are capitalized and others aren't ("chair" vs "Desk").

JRegimbal (changes requested):
Runs fine locally, some necessary changes in the code. There's also no CI/CD for it, so please make one.

preprocessors/object-detection-azure/objdetect.py
gp1702: Isn't that the name for the YOLO model?

preprocessors/object-detection-azure/objdetect.py
gp1702: I don' think we can get that information from azure.

preprocessors/object-detection-azure/objdetect.py
gp1702: Is there a reference for the taxonomy of error codes ?

preprocessors/object-detection-azure/objdetect.py
gp1702: Same as below, if there's a document with referencing to common error codes, please share.

preprocessors/object-detection-azure/objdetect.py
gp1702: For the CI CD, please share documentation of how to make one.

preprocessors/object-detection-azure/objdetect.py
JRegimbal: Return 500 if there's no API key and include a log message that there's no key. This just throws a KeyError as is.

gp1702: What does 500 stand for? . There can be many issues with the key, the file can be missing or the key might be outdated. Does this error code handle all such cases ?

.github/workflows/object-detection-azure.yml
```diff
@@ -0,0 +1,82 @@
+name: Graphic Tagger
```
JRegimbal, line 1: Name

.github/workflows/object-detection-azure.yml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          flavor: |
+            latest=${{ env.TAGGED }}
+          tags: |
+            type=match,enable=${{ env.TAGGED }},priority=300,pattern=preprocessor-object-detection-azure-(\d+.\d+.\d+),group=1
+            type=raw,priority=200,value=unstable
+            type=raw,priority=100,value=${{ env.timestamp }}
+          labels: |
+            org.opencontainers.image.title=IMAGE Preprocessor Graphic Tagger
```
JRegimbal, line 70: Name and description here.

preprocessors/object-detection-azure/objdetect.py
JRegimbal: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500 It's an internal server error. You can specify what happened in the log.

JRegimbal (changes requested):

preprocessors/object-detection-azure/objdetect.py
gp1702: Still unclear, more details needed.

JRegimbal: When our quota was reached, the request to Azure stopped returning data. If Azure becomes unreachable for a time, we don't get a 200 (OK) response. In these cases, what would happen here is the preprocessor would report nothing in logs, return 204 (No Content, what we use to say everything is okay and we are returning no data) and exit. In these cases, something has went wrong and if nothing is done to report it here we won't be able to diagnose the problem.

docker-compose.yml
```diff
@@ -88,6 +88,13 @@ services:
                     devices:
                     - driver: nvidia
                       capabilities: ["gpu", "utility", "compute"]
+    object-detection-azure:
+        image: ghcr.io/shared-reality-lab/image-preprocessor-object-detection-azure:unstable
+        labels:
+            ca.mcgill.a11y.image.preprocessor: 6
```
JRegimbal, line 94: Run earlier, otherwise sorting/grouping won't apply.
```suggestion
            ca.mcgill.a11y.image.preprocessor: 3
```

JRegimbal (changes requested):

docker-compose.yml
gp1702: But 3 is already assigned to the YOLO, does this have to run before grouping and sorting ? in that case it would be better to change their numbers.

jeffbl: (for example, OCR and autour are both labeled `2` since they can run in parallel - they are completely independent of each-other.)

preprocessors/object-detection-azure/objdetect.py
```diff
@@ -0,0 +1,194 @@
...
+                int(response.headers['content-length']) == 0:
+            return "Invalid response from azure"
+        elif 'content-type' in response.headers and \
+                isinstance(response.headers['content-type'], str):
+            if 'application/json' in response.headers['content-type'].lower():
+                if response.content:
+                    result = response.json()
+                    label = process_results(result)
+                else:
+                    return "Response content missing"
```
JRegimbal, line 108: These return statements without an explicit code default to 200 OK as explained [in the Flask docs](https://flask.palletsprojects.com/en/2.0.x/quickstart/#about-responses). Please change these to return an appropriate 500 code.

JRegimbal (changes requested):

preprocessors/object-detection-azure/objdetect.py
```diff
@@ -0,0 +1,194 @@
...
+        add.format(region),
+        data=image,
+        headers=headers,
+        params=params
+    )
+    if response.status_code == 200 or response.status_code == 201:
+
+        if 'content-length' in response.headers and \
+                int(response.headers['content-length']) == 0:
+            return "Invalid response from azure", 204
```
JRegimbal, line 100: These are codes that would be (ideally) returned by Azure. Not by us. In these cases we're experiencing a failure. These should be 5xx error codes, probably just 500.

JRegimbal (changes requested):

JRegimbal (approved):
Seems to be working correctly when tested locally. I'll merge in a bit if nobody else has any concerns.

jeffbl:
I won't be able to test beyond superficial until weekend, but this is an isolated preprocessor we can just turn off if problems, so please merge and deploy to unicorn for testing. This may help us significantly if we need to offload pegasus during CSUN, so it would be good to have that option.

JRegimbal:
Alright. I'll merge and deploy on Unicorn. Keep in mind that this means the YOLO model will not be running there.

PR #386, opened 2022-03-04
reduce OCR fetch wait time 1s -> 10ms

BenMacnaughton:

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/preprocessors/ocr/ocr.py
+++ b/preprocessors/ocr/ocr.py
@@ -144,7 +144,7 @@ def analyze_image(source):
         if time.time() - start_time > 3:
             logging.error("Azure request timed out")
             return None
-        time.sleep(1)
+        time.sleep(0.01)
 
     # Check for success
     if read_result.status == OperationStatusCodes.succeeded:


BenMacnaughton:
Single line to change the time spent waiting between loops when fetching the Azure Read API results

preprocessors/ocr/ocr.py
```diff
@@ -144,7 +144,7 @@ def analyze_image(source):
         if time.time() - start_time > 3:
```
jaydeepsingh25, line 144: Why do we need to call get API multiple times in a loop?
Any specific reason for this?

BenMacnaughton: since two requests are made, one to create a OCR record and one to fetch that record, we have to query the location of the record until it is created. until then it may return "in progress" in which case we have to continue waiting for its creation

aidanwilliams09: Should we even have a wait in there? Is there a downside to querying constantly in terms of price or timeouts?

jeffbl:
A busy loop with no delay will do something like peg a CPU core and potentially cause other issues. The correct fix would be to do something asynchronous that calls you back when it is done. It would be very unusual for a modern web API not to have this as an option. Worth checking documentation, and pinging @jaydeepsingh25 if you can't find an alternative.

If a loop like this is necessary, it must have a delay to avoid hammering the server - there is no way it is going to reply a million times each second. Given user latency, I'd say a decent timeout value would be 200ms. On average, user will then have a delay of ~100ms, which is not terrible, and only queries the server 5x/s (which still seems terrible to me, but...)

BenMacnaughton:
There doesnt seem to be a workaround provided by MS here. The strategy I employed comes from their suggested implementation documents

JRegimbal:
Hi @BenMacnaughton, it seems like while the `read` API endpoint does have you poll to check for completion, it's also meant to support documents hundreds or thousands of pages long. The [`ocr` API endpoint](https://westus.dev.cognitive.microsoft.com/docs/services/computer-vision-v3-2/operations/56f91f2e778daf14a499f20d) seems to be synchronous and meant for shorter documents we'd likely encounter for IMAGE.

BenMacnaughton:
Ok thanks for looking into that! I'll make sure it's feasible to make the change. Hopefully this will also increase our recognition range because I've noticed that the read API can fail on some seemingly trivial examples

BenMacnaughton:
@JRegimbal the OCR API seems to provide regions of text that contain individual lines. https://westus.dev.cognitive.microsoft.com/docs/services/computer-vision-v3-2/operations/56f91f2e778daf14a499f20d
It might make the most sense to have the preprocessor return these regions as their lines are likely to be contextually related. Would you agree?

JRegimbal:
Hard to say without seeing the results in practice, but it seems like a reasonable thing to try.

PR #403, opened 2022-03-16
change bounding box values to normalized

BenMacnaughton:

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/preprocessors/ocr.schema.json
+++ b/preprocessors/ocr.schema.json
@@ -20,12 +20,13 @@
                     "bounding_box": {
                         "type": "array",
                         "title": "Bounding box",
-                        "description": "The bounding box of the text in pixels: [X-TL, Y-TL, X-TR, Y-TR, X-BR, Y-BR, X-BL, Y-BL]",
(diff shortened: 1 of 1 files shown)


BenMacnaughton:
This has been tested on Unicorn with a number of graphics requests containing text.

BenMacnaughton:
The only change here is to convert the OCR text bounding box to a normalized set of coordinates so that it can be compared against normalized coords from other preprocessors

BenMacnaughton:
Yes this is just to do with the differences between the OCR api and the Read api. (The former returns 2 points always forming a rectangle whereas the latter returns 4 points that can for some arbitrary quadrilateral). Even when we were using the Read api I was not making use of these additional points so I think it's fine

JRegimbal (approved):

JRegimbal:
@BenMacnaughton the PR to the OCR API is in a ready or near-ready state, right?

BenMacnaughton:
@JRegimbal yes ready as far as I am concerned

PR #406, opened 2022-03-21
Switch to mp3

### LINKED ISSUE #405: Investigate formats smaller than FLAC to reduce bandwidth/latency
We [switched to flac](https://github.com/Shared-Reality-Lab/IMAGE-server/pull/401) from wav to make this significantly better, but we can probably reduce the size of downloads even further with a format like MP3, assuming that it does not degrade spatialization/quality. We may have users on DSL lines, which are about 1MB/s download. That means that a 6MB rendering can take 6s to load, which effectively doubles the total latency. Note that this also would reduce the delay in the [progress status being incorrect](https://github.com/Shared-Reality-Lab/IMAGE-browser/issues/205), but that should still be resolved independently.

JRegimbal:
Resolve #405. Note that SuperCollider does not natively support MP3 and the [MP3 quark](https://github.com/supercollider-quarks/MP3) doesn't seem to provide an easy way to work with NRT. Instead, we're just calling [LAME](https://lame.sourceforge.io/) directly, which is what the quark would do under the hood anyway.

This involves changing writing and adding a conversion step in SuperCollider and changing from FLAC to MP3 output files in the respective handlers. The process was tested with the photo audio handler ~. The others should just work with the same changes.~, the autour handler, and the high charts handler.

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/handlers/autour-handler/src/server.ts
+++ b/handlers/autour-handler/src/server.ts
@@ -163,7 +163,7 @@ app.post("/handler", async (req, res) => {
         scData["ttsFileName"] = inFile;
         jsonFile = filePrefix + Math.round(Date.now()) + ".json";
         await fs.writeFile(jsonFile, JSON.stringify(scData));
-        outFile = filePrefix + uuidv4() + ".flac";
+        outFile = filePrefix + uuidv4() + ".mp3";
         await fs.writeFile(outFile, "");
         await fs.chmod(outFile, 0o664);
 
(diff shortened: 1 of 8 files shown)


JRegimbal:
Assigning to @jeffbl just for approval on the approach here.

jeffbl:
Approach seems fine. I don't see a specification for the MP3 bitrate? Is it using a default, and if so, what is it?

JRegimbal:
I have it set to variable bitrate (`-v` flag) since that seemed like a sensible default to try.

jeffbl:
Got it, thanks. Looks like it defaults to max. quality (although man page for lame indicates otherwise, so not clear?), so we'll have to keep an eye on how much it reduces file size. To adjust, it looks like the `-V` flag lets VBR be adjusted for quality: https://lame.sourceforge.io/vbr.php

PR #445, opened 2022-06-14
Initial commit for SVG visualization schema.

### LINKED ISSUE #442: Create new schema for SVG visualization functionality
To implement #117, create new schema required for the [new rendering type](https://github.com/Shared-Reality-Lab/IMAGE-browser/issues/229) to be implemented in the extension.

JRegimbal:
An SVG visualization schema to resolve #442.

In brief, the current implementation is called `ca.mcgill.a11y.image.renderer.SVGLayers`. It is an object with two keys: `layers`, which is an array containing objects that include a `label` for the layer and an `svg` encoding the SVG as a data URL; and `description` which optionally provides a description of the kind of information returned across the layers. I.e., an explanation of the preprocessor and any non-obvious visualization decisions made for its output data.

Tests aside from schema compilation have not been run at this time.

### DIFF
--- a/renderers/svglayers.schema.json
+++ b/renderers/svglayers.schema.json
@@ -1,0 +1,39 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://image.a11y.mcgill.ca/renderers/svglayers.schema.json",
+    "title": "Layered SVG Renderer",
+    "type": "object",
+    "description": "Overlay visualization of data from an input graphic, possibly containing multiple layers.",
+    "definitions": {
+        "svgLayer": {
+            "description": "A single SVG graphic encoded as a data URL.",
+            "type": "string",
+            "pattern": "^data:image/svg\+xml.+$"
+        }
+    },
+    "properties": {
+        "layers": {
+            "description": "The layer(s) representing data. It should be possible to overlay each layer on the original graphic to meaningfully convey the desired information.",
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "svg": { "$ref": "#/definitions/svgLayer" },
+                    "label": {
+                        "type": "string",
+                        "maxLength": 300,
+                        "description": "A short, descriptive label for the SVG in this layer."
+                    }
+                },
+                "required": [ "svg", "label" ]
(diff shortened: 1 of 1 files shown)


renderers/svglayers.schema.json
```diff
@@ -0,0 +1,39 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://image.a11y.mcgill.ca/renderers/svglayers.schema.json",
+    "title": "Layered SVG Renderer",
+    "type": "object",
+    "description": "Data for verlay visualization of data from an input graphic, possibly containing multiple layers. ca.mcgill.a11y.image.renderer.SVGLayers.",
```
jeffbl, line 6:
```suggestion
    "description": "Data for overlay visualization of data from an input graphic, possibly containing multiple layers. ca.mcgill.a11y.image.renderer.SVGLayers.",
```

renderers/svglayers.schema.json
```diff
@@ -0,0 +1,39 @@
...
+        "layers": {
+            "description": "The layer(s) representing data. It should be possible to overlay each layer on the original graphic to meaningfully convey the desired information.",
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "svg": { "$ref": "#/definitions/svgLayer" },
+                    "label": {
+                        "type": "string",
+                        "maxLength": 300,
```
jeffbl, line 24: Assuming this is the string that will be displayed in the UI, in a dcropdown, 300 is probably too many?

jeffbl (approved):

renderers/svglayers.schema.json
JRegimbal: Probably too many, but I'm more worried about reasonable labels being rejected for going slightly over a limit than them ending up with someone trying to put in a novella! Do you have a suggestion to set it to?

jeffbl: Hm. This is primarily for debugging, so it can be longer than something I'd consider end-user UI. Looking at current window, even if dropdown is all the way across the screen, I'd say 200 is more than enough, and 100 would enforce more discipline in naming. 150?

Issue #460, opened 2022-06-30
Create Schemas for additional ML models


rohanakut:
In order to create the preprocessors mentioned in #455 we need to create schemas for output validation. This issue creating a schema for following preprocessors:

- [ ] Action recognition preprocessor (PR - #459 )
- [ ] Emotion recognition preprocessor (PR - #461 )
- [ ] Dominant Color Detector
- [ ] Celebrity Recognition
- [ ] Landmark Detection

rohanakut:
@jeffbl closing this, since it was decided that the schemas would not be created for the current implementation

PR #461, opened 2022-06-30
Emotion recognition schema

rohanakut:
This PR is wrt #460. The schema is created for emotion recognition models. The following is the expected output from the schema

```
{
  "emotion_age_gender": [
    {
      "ObjectID": 1,
      "possible_emotion": [
        {
          "emotion": "happy",
          "confidence": 0.99
        },
        {
          "emotion": "neuutral",
          "confidence": 0.59
        }
      ],
      "possible_age": 10,
      "possible_gender": "Woman"
    },
    {
      "ObjectID": 2,
      "possible_emotion": [
        {
          "emotion": "sad",
          "confidence": 0.99
        },
        {
          "emotion": "neutral",
          "confidence": 0.59
        }
      ],
      "possible_age": 10,
      "possible_gender": "Man"
    },
  ]
}
```

**Note:** The age and gender are the attributes supplied by the preprocessor. However, they are notorious to perform poorly. Hence they have not been mentioned in the required field


JRegimbal (commented):
More information is needed. Where is the vocabulary coming from? Which models are you using to implement? Is the object ID from YOLO? Is there anything done to ensure that every person treated by this model/these models was detected by YOLO, and otherwise how is ID handled? Is confidence as reported here actually useful or are we running into the famous problem we encountered previously with it as a metric? Is there any reason to report age and gender if they're notoriously inaccurate? I assume "gender" is just being reported as either male/female, which would already be a problematic set of outputs to work with aside from accuracy problems. I strongly recommend dropping gender as a reported feature, even an optional one.

rohanakut:
The vocabulary is produced by the [DeepFace](https://github.com/serengil/deepface) model. There are a few conditions which need to be satisfied before the preprocessor is run. The following are the conditions:
1. **A person needs to be present in the Image:** This person will be detected by YOLO model.
2. **A check on number of people in the image and the % of image occupied by these people:** This check would be done based on the bounding boxes produced by the YOLO model. If there are a large number of people in the image and the bounding boxes are very small in size then this model would not be run

Thus in a nutshell, this model would only be run when the YOLO model would detect people. In addition, the model would run only on images containing small number of people with these people occupying a large portion of image.

For example, the following image would be an ideal image for this preprocessor. This image has been referred form [CTV news](https://www.ctvnews.ca/world/u-s-capitol-riot-more-people-turn-up-with-evidence-against-donald-trump-1.5972979)
Screen Shot 2022-07-04 at 10 48 24 AM

JRegimbal:
Thanks, I'll look into DeepFace more and especially the labels it uses.

1. If you're confident we won't have mismatches (e.g., YOLO detects a person that this misses, YOLO misses a person this detects) I think that it is fine to proceed. I would try and test this out in practice though early on and plan out a solution for edge cases that, however unlikely, are still possible and will happen given enough people sending enough images with no restriction on source material.
2. Fair enough, please include documentation to that effect in the schemas or point to a place that explains how it can and should be used.
3. In short, misclassifications can be harmful, a binary male-female label categorically excludes nonbinary people, and systems are disproportionately harmful to transgender people broadly. [This paper from CHI'18 goes into the HCI implications in more detail.](https://dl.acm.org/doi/abs/10.1145/3173574.3173582) I don't know what specific problem you're trying to solve by including gender, but I'm skeptical it's worth including. Personally, I would not be comfortable helping to implement or maintain such a functionality.

Cybernide:
1. Ah, it seems that I misunderstood and assumed that "emotion" referred to the global characteristics of a graphic. I see now that you're trying to get the emotions of people in the picture. I would also prefer to have congruence between the people/object detections, but I think that this is something that we should investigate and maybe there's a way to handle it (as in, we can use the emotion detector as a different interpretation as Jeff once suggested? Possibly? When we did our structured interviews on photos, participants were absolutely interested in emotions depicted, so I'm not entirely convinced that we can't use a different model from YOLO, we just might have to be careful in how we show the results of the models, if that makes sense.

3. Without going into a whole lot of talk about social and cultural implications of gender here on GitHub, we do not want to be using anything that bins people into the exclusive categories of "male" and "female". @rohanakut I don't have a lot of time this week for face to face discussions, but if you're having trouble with this, send me a message on Slack and we can talk.

rohanakut:
@JRegimbal I understand what you are trying to say. I will remove the age and gender part as that might hurt sentiments of a few users. Apart from that are there any other concerns for the PR?

jeffbl:
@rohanakut Is this specifically for your thesis, or for eventual production deployment? If you have a specific use case that is vetted for your thesis, which can be done without deployment, then the bar is lower. However, note that (for example) [Microsoft recently removed emotion recognition from their Azure toolkit](https://www.theverge.com/2022/6/21/23177016/microsoft-retires-emotion-recognition-azure-ai-tool-api) since it basically doesn't work, and is subject to many of the sorts of issues discussed above. We'd have to have a broader meeting to vet this before it would be deployable, given then known issues and harms with some models. It also depends on how we couch the results, to emphasize they can be wildly wrong, especially if we don't have any sort of usable confidence metric.

rohanakut:
@jeffbl initially this was planned for my thesis but I am open to keeping this in the production. I now understand that emotion recognition could be a debatable topic but I think we should first vet the type of output generated by this model. When I personally tested the emotion recognition model on 10 images, I felt that that the output was not controversial. The generated output might not have 100% agreement with humans, since the emotions are subjective, but I did not find anything controversial in the outputs that I generated.

@JRegimbal I agree that gender and age could be controversial. I will remove that in the next commit.

rohanakut:
closing this PR as it was decided that schemas are not needed for the implementation

PR #462, opened 2022-07-04
Deployment information

### LINKED ISSUE #449: Create documentation for bringing up a new IMAGE server
Although all of the pieces are present to do so, we do not have a set of instructions specific to bringing up an IMAGE server from scratch. This would be useful to:
1. Us, if we ever need to do it if one of our servers goes kaput
2. Developers who want to run their own test server instance for debugging/testing
3. External groups wanting to run their own server

Note that this should not have duplicate/overlapping content with specific README files in the repo, e.g., #448

JRegimbal:
Add a `DEPLOYMENT.md` file to resolve #449. It is currently not a step-by-step summary, but aims to explain how the pieces fit together to create our testing and deployment setups.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -17,6 +17,12 @@ we have practical experience with. You should have installed:
     - Many of these containers do not use or do not need to use a GPU, but performance will be greatly reduced.
 - [Git](https://git-scm.com/)
 
+Note that running every GPU-ready container on a GPU will use multiple GB of memory
+on the device. On many GPUs, an out-of-memory error will occur. For most GPU-ready
+containers, switching it to CPU is a matter of removing the deployment lines
+specifying it should have access to a GPU. Otherwise, there will be an environment
+variable specifying which to use (e.g., "cuda", "cpu").
+
 ## Unstable vs Latest
 
 Our Docker images our tagged in four possible ways:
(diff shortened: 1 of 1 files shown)


DEPLOYMENT.md
```diff
@@ -39,6 +39,8 @@ git clone --recurse-submodules git@github.com:Shared-Reality-Lab/IMAGE-server.gi
 The default compose file within there (`docker-compose.yml`) is a useful base for testing. Each image will be downloaded from the Github Container Registry
 using the "unstable" tag.
 
+This can be brought up using `docker-compose up -d`.
+
 ## For Production
 
 For production, it is recommended to use "latest" and limit the number of running services for additional stability.
```
jeffbl, line 46: suggest: "and limit the number of running services to fit your CPU, memory, and GPU limitations, for stability"

jeffbl (approved):
Looks good! Two overall suggestions:
1. Adding few words about resource requirements including possibility of requiring a GPU, and that some are configurable to run on either CPU or GPU. (Generic, just raising the issues, not specific to preprocessors/handlers, since specifics should be in those specific README's....)
2. Reword the Azure/Google places key sections to be a bit more generic, so that if we add other such services, this is generic enough. If there are specifics, could they be comments in the docker-compose.yml directly instead, and we just point to this as an issue in this file? Here, they could be pointed to as examples, rather than a complete list?

Issue #429, opened 2022-04-22
Implement additional measures/interpretations to raw chart data


jeffbl:
@florian-grond has suggested that for sonifying charts, having information such as maxima and minima already calculated would be useful, and would probably help people building other handlers as well. There is some debate about whether this should be an additional preprocessor that ingests chart data and adds to it, or a more dynamic service/library called by the handlers. Ideas include things like:

- labeling each data point with its local maximum / minimum window, e.g., how many points before and after are lower/higher than it. (and/or some other max/min metric)
- - things specific to charts like candlestick charts, if we need to identify common patterns ourselves.
- @florian-grond other examples?

Cybernide:
For your first bullet, it seems as if we would need to implement a way of navigating from "window" to "window" - should I make this a UX/UI issue?

jeffbl:
@Cybernide that is between you and the people designing the handler(s) for this, which means @florian-grond / @JRegimbal. Design work for this is in #431 .

jeffbl:
After discussion today, first set of new columns will be these four:
1. number of contiguous points to right that are higher
2. number of contiguous points to the left that are higher
3. number of contiguous points to the right that are lower
4. number of contiguous points to the left that are lower

We verified with @florian-grond that this should be sufficient for the initial sonfications he wants to try.

For intial PR, brute force is ok, but there are easy optimizations based on previous point that should make this much faster than the worst case O(n^2), which could bog down on charts with many points. For example, if the point to the left of the one currently under consideration had 15 points lower to the left, and the current point is higher than the one to the left, re-counting is not necessary since we know that there are 15+1=16 points lower to the left.

Once this is implemented and checked in (this is the first functionality requiring an actual charts preprocessor to be implemented), doing further tweaks/columns should be pretty trivial, barring CPU load for the actual processing.

Discussed adding some sort of moving average column as well, which might be useful in tandem with the data for finding local max/min values for some of Florian's use cases.

jeffbl:
@jaydeepsingh25 This should be closed by #475, correct? Any further columns to be implemented should be logged as new work items.

jaydeepsingh25:
yes, the issue is closed by #475

PR #474, opened 2022-08-10
Charts Preprocessor schema

jaydeepsingh25:
The PR adds the schema for charts preprocessor data (Issue: https://github.com/Shared-Reality-Lab/IMAGE-server/issues/429)

## Required Information

## Pull Request Checklist

- [ ] The Docker image builds

### DIFF
--- a/preprocessors/charts.schema.json
+++ b/preprocessors/charts.schema.json
@@ -1,0 +1,51 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://image.a11y.mcgill.ca/preprocessors/charts.schema.json",
+    "type": "object",
+    "title": "Charts preprocessor",
+    "description": "Extracts additional information from the datapoints present in the graph",
+    "properties": {
+        "lines": {
+            "type": "array",
+            "title": "dataPoints",
+            "description": "The dataPoints in the graph",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "x": {
+                        "type": "number",
+                        "title": "x",
+                        "description": "x axis value"
+                    },
+                    "y": {
+                        "type": "number",
+                        "title": "y",
+                        "description": "y axis value"
+                    },
+                    "lowerPointsOnLeft": {
+                        "type": "number",
+                        "title": "Lower points on left",
+                        "description": "Contigous points lower than the data Point, lying on the left side on the x axis"
+                    },
+                    "higherPointsOnLeft": {
+                        "type": "number",
+                        "title": "higher points on left",
+                        "description": "Contigous points higher than the data Point, lying on the left side on the x axis"
+                    },
+                    "lowerPointsOnRight": {
+                        "type": "number",
+                        "title": "Lower points on right",
+                        "description": "Contigous points lower than the data Point, lying on the right side on the x axis"
+                    },
+                    "higherPointsOnRight": {
+                        "type": "number",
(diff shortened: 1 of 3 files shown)


preprocessors/charts.schema.json
```diff
@@ -0,0 +1,51 @@
...
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://image.a11y.mcgill.ca/preprocessors/charts.schema.json",
+    "type": "object",
+    "title": "Charts preprocessor",
+    "description": "Extracts additional information from the datapoints present in the graph",
+    "properties": {
+        "dataPoints": {
+            "type": "array",
+            "title": "dataPoints",
+            "description": "The dataPoints in the graph",
```
JRegimbal, line 11: `title` and `description` are purely annotations. I'm not sure why you write it as "dataPoints", does this have a special meaning elsewhere?

preprocessors/charts.schema.json
```diff
@@ -0,0 +1,51 @@
...
+                    },
+                    "y": {
+                        "type": "number",
+                        "title": "y",
+                        "description": "y axis value"
+                    },
+                    "lowerPointsOnLeft": {
+                        "type": "number",
+                        "title": "Lower points on left",
+                        "description": "contiguous points lower than the data Point, lying on the left side on the x axis"
```
JRegimbal, line 28: This mention of "contiguous" is odd in context with the name and title not specifying any kind of range. This is the count of points to the left of the current point before a point is reached with an equal or higher value? Is this count actually useful if points aren't equally spaced on the x axis? Identical questions and concerns for the other higher/lower points on left/right keys.

preprocessors/charts.schema.json
```diff
@@ -0,0 +1,51 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://image.a11y.mcgill.ca/preprocessors/charts.schema.json",
+    "type": "object",
+    "title": "Charts preprocessor",
+    "description": "Extracts additional information from the datapoints present in the graph",
```
JRegimbal, line 6: This only has the possibility of encoding information on a graph with two quantitative axes...please update the description/title/ID to make this name more accurate to what is encoded or, if this is meant to be a generic data type that will be extended to meet a larger variety of graphs, include a mention that this is not complete.

preprocessors/charts.schema.json
JRegimbal: More questions...what does "lying on the left side on the x axis" mean here? I've been assuming left of the point but this isn't actually stated anywhere. Is this to the left of where the y axis was drawn on the screen? Is left/right defined some other way?

JRegimbal (commented):
Structurally looks fine, but clarification needed. Mainly I'm confused on the scope of this schema file. This is named as *the* chart schema, but assumes a numerical x-y plot and only outputs information that makes sense in that context. Is this meant to be expanded so that it supports more? If so, I'm worried about how generic the name "dataPoints" is considering how narrowly scoped its function is. If not, then the name needs to better fit what this is.

Also I'm a bit confused by the higher/lower points on the right/left entries and would expect others reading this to have similar problems. If I were to read this with no knowledge from being on the team, I would probably not know what information I was being given or how I could use it.

JRegimbal (approved):
Assuming these numbers are requested by @florian-grond and there's a plan for them...approving. Thanks!

PR #491, opened 2022-08-28
Improve OSM streets description & scalability

samueljohnsegun148:
Initial steps to address having too many nameless streets, and to introduce map scalability.

## Required Information

## Pull Request Checklist

- [ ] The Docker image builds

### DIFF
--- a/preprocessors/openstreetmap/osm_service.py
+++ b/preprocessors/openstreetmap/osm_service.py
@@ -43,9 +43,8 @@ def server_config1(url, bbox_coord):
     street_data = api.query(
         f"""
     way({lat_min},{lon_min},{lat_max},{lon_max})[highway];
-
     (._;>;);
-    out geom;
+    out body;
     """
     )
     return street_data
@@ -123,21 +122,26 @@ def process_streets_data(OSM_data):
                 oneway = bool(oneway)
             else:
                 oneway = oneway
-                way_object = {
-                    "street_id": int(way.id),
-                    "street_name": way.tags.get("name"),
-                    "street_type": way.tags.get("highway"),
-                    "addr:street": way.tags.get("addr:street"),
-                    "surface": way.tags.get("surface"),
-                    "oneway": oneway,
-                    "sidewalk": way.tags.get("sidewalk"),
-                    "maxspeed": way.tags.get("maxspeed"),
-                    "lanes": lanes,
-                    "nodes": node_list,
-                }
-                # Delete key if value is empty
-                way_object = dict(x for x in way_object.items() if all(x))
+
+            way_object = {
+                "street_id": int(way.id),
+                "street_name": way.tags.get("name"),
+                "street_type": way.tags.get("highway"),
+                "addr:street": way.tags.get("addr:street"),
+                "surface": way.tags.get("surface"),
+                "oneway": oneway,
+                "sidewalk": way.tags.get("sidewalk"),
+                "maxspeed": way.tags.get("maxspeed"),
+                "lanes": lanes,
+                "nodes": node_list,
+            }
+            # Delete key if value it is empty
+            way_object = dict(x for x in way_object.items() if all(x))
+
+            # Include only streets with names
+            if "street_name" in way_object:
                 processed_OSM_data.append(way_object)
+            processed_OSM_data.append(way_object)
     except AttributeError:
         error = 'Overpass Attibute error. Retry again'
         logging.error(error)
@@ -394,9 +398,10 @@ def enlist_POIs(processed_OSM_data1, amenity):
                 # check if "cat" key is in the node
                 if key_to_check in nodes[node]:
                     if nodes[node]["cat"]:  # ensure the "cat" key has a value
-                        POIs.append(nodes[node])
+                        # Check to remove duplicate intersections
+                        if nodes[node] not in POIs:
+                            POIs.append(nodes[node])
     if amenity is not None and len(amenity) != 0:
-        # POIs = [objs for objs in amenity]
         for objs in range(len(amenity)):
             POIs.append(amenity[objs])
     return POIs  # POIs is a list of all points of interest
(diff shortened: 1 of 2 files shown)


preprocessors/openstreetmap/osm_service.py
```diff
@@ -123,20 +122,27 @@ def process_streets_data(OSM_data):
...
+            }
+            # Fetch as many tags as possible
+
+            way_object["nodes"] = node_list
+
+            # Delete key if value is empty
+            way_object = dict(x for x in way_object.items() if all(x))
+
+            # Include only streets with names
+            if "street_name" in way_object:
```
JRegimbal, line 145: I'm still skeptical that we should drop all nameless paths from the output. Yes, this is useful for Florian's output, but restricts how we can use the data later on. Since this is so minor, keep it in now, but we may have to remove these lines.

preprocessors/openstreetmap/osm_service.py
```diff
@@ -394,9 +400,10 @@ def enlist_POIs(processed_OSM_data1, amenity):
                 # check if "cat" key is in the node
                 if key_to_check in nodes[node]:
                     if nodes[node]["cat"]:  # ensure the "cat" key has a value
-                        POIs.append(nodes[node])
+                        # Check to remove duplicate intersections
+                        if nodes[node] not in POIs:
```
JRegimbal, line 404: Excellent!

preprocessors/openstreetmap/osm_service.py
```diff
@@ -543,3 +550,14 @@ def get_coordinates(content):
     """
     if 'coordinates' in content.keys():
         return content['coordinates']
+
+
+def scale_distance(content):
```
JRegimbal, line 555: It looks like you're adding distance to the request parameters...first off, you need to specify units for this in the code. I'm assuming this is in meters, which would be perfect. Second, adding new fields to the request from the extension is much more involved than just this...currently this wouldn't be valid and I don't believe we can even get this data from the embedded maps at this point. I'll get discussion on this added to the next tech arch meeting if not before then. In the meantime, with this feature we'll need to be very careful in how we choose the default value (here, I believe this would be within a square of 400 m since the sides are each 200 m from the center, right?) since this default is what will end up being used for the foreseeable future.

JRegimbal (commented):
This is looking very good! There are some points that require larger discussion, but everything here is looking like it's on the right track. Would it be possible to include some features to merge streets with different IDs and the same names? Often these have "intersections" where one street ends and immediately, from the same node, continues on with a new ID. At least, this seems to be common in the areas I've checked in Quebec, but I suspect that this isn't just a quirk of the local maps!

preprocessors/openstreetmap/osm_service.py
samueljohnsegun148: Yes, the distance is in metres, with the map area being 400sqm as rightly said. Your concerns are well noted.

preprocessors/openstreetmap/osm_service.py
samueljohnsegun148: > I'm still skeptical that we should drop all nameless paths from the output. Yes, this is useful for Florian's output, but restricts how we can use the data later on. Since this is so minor, keep it in now, but we may have to remove these lines.

Ok

jeffbl:
@samueljohnsegun148 please assign to Juliette when ready for review again

JRegimbal:
Followup RE merging streets with the same name and different IDs as mentioned in today's meeting: this may occur because some characteristic of the street (e.g., number of lanes) changes. While this isn't relevant to renderings we are currently thinking of, it may be relevant later. Street merging should not be done in the preprocessor and should be done opportunistically in the handler.

samueljohnsegun148:
> Followup RE merging streets with the same name and different IDs as mentioned in today's meeting: this may occur because some characteristic of the street (e.g., number of lanes) changes. While this isn't relevant to renderings we are currently thinking of, it may be relevant later. Street merging should not be done in the preprocessor and should be done opportunistically in the handler.


Yes, I quite agree. Nevertheless, my personal concern with merging the streets of the same name in the preprocessor is that it may lead to loss of information, though not in most cases. In OSM, streets are usually split into segments when attributes change. A typical OSM way is split into smaller segments ending up having the same name (with each having a unique ID) mostly because of different tags on the segments, such as speed limits, restrictions, parking conditions, or even split to accommodate different bus routes, etc. So, attempting to join these segments may lead to losing this information, though that may not be in all cases.
Merging is okay. It however depends on what level of information we are willing to sacrifice, that would not necessarily impact our overall objectives.
The idea of doing the merging in the handler is also a good one, even though I don’t really have much idea on how the handler works internally. However, I’m not honestly too sure if the handler can easily handle the geometric manipulations that may be involved in the merging. I stand to be corrected. But what I'm trying to say is that, it is not just about joining the streets, but such that the merging is done with all the points (nodes) orderly connected to form a single street as it would be on the map itself. For instance, if there are four segments with the same name, some kind of “sorting” is required to determine which segment comes first, second, and so on, to have the orderly arrangement of the segments as one. This sounds so simple, but may be complex as well.

JRegimbal:
Yeah, I'd advise against any merging in the preprocessor unless we start seeing cases where all information *except* for the IDs are the same. And I'm in complete agreement with you that the complexity of it in a handler may get a bit out of hand, but so long as we don't have multiple handlers doing the same transformations (which would justify a new preprocessor) it's the "right" way to do things.

No work on way merging needs to be done for now and I suppose we'll deal with it when we need to. The only ways that would definitely really benefit from this merging in a preprocessor in the examples I've tried are footpaths, but if I remember correctly you've already filtered these out anyway.

samueljohnsegun148:
Yeah, exactly.

samueljohnsegun148:
This is ready for a review now.

jeffbl:
@samueljohnsegun148 I'm assigning to @JRegimbal - reviewers will not look at a PR until it is directly assigned to them.

JRegimbal (approved):

JRegimbal:
Merging, however looking at the output duplicate intersections are *not* being removed. The check on line 404 does not exclude several intersections that are identical based on node ID (which is what's important) and differs only in terms of description, where one is "Street A intersecting Street B" and the other is "Street B intersecting Street A".

This will be split out into another issue.

Issue #507, opened 2022-10-20
Remove Duplicate Intersections in OSM Output


JRegimbal:
Duplicate intersections exist in OSM preprocessor output where entries are equivalent in terms of node ID, latitude, and longitude, but have slightly different `name` fields. For example the following are returned as separate entries in a `points_of_interest` field:

```json
{
    "id": 540192036,
    "lat": 46.4920905,
    "lon": -72.9606394,
    "cat": "intersection",
    "name": "Avenue Principale intersecting Rue Joseph-Garand"
}
```
```json
{
    "id": 540192036,
    "lat": 46.4920905,
    "lon": -72.9606394,
    "cat": "intersection",
    "name": "Rue Joseph-Garand intersecting Avenue Principale"
}
```

samueljohnsegun148:
Okay thanks @JRegimbal. This duplicate type in particular was deliberately made to be so, with the idea to convey which street is intersecting the other based on the street being followed. For instance, if passing  through 'street A', and A is being intersected by street ' B', then the presentation would be A intersects B, and vice-versa. So the idea is that since A is named first, one would know that the current street being transversed is A and not B. However, now that it's clear that such would not be necessary, I will simply remove that to avoid the duplicate issue raised in a new PR. Thanks

PR #509, opened 2022-10-21
Solve Duplicate Intersections in OSM and Nodes returned outside of the expected bounds

samueljohnsegun148:
This is to address the issue of duplicate intersections in OSM in #507 and also the issue of nodes being returned by OSM Preprocessor outside of the expected bounds in #510.

## Required Information

## Pull Request Checklist

- [x] The Docker image builds

### DIFF
--- a/preprocessors/openstreetmap/osm_service.py
+++ b/preprocessors/openstreetmap/osm_service.py
@@ -387,7 +387,8 @@ def get_amenities(bbox_coord):
 def enlist_POIs(processed_OSM_data1, amenity):
     # Keep all identified points of interest in a single list
     POIs = []
-    nodes_id = []
+    nodes_ids = []
+    unique_POIs = []
     if len(processed_OSM_data1):
         for obj in range(len(processed_OSM_data1)):
             nodes = processed_OSM_data1[obj]["nodes"]
@@ -399,10 +400,17 @@ def enlist_POIs(processed_OSM_data1, amenity):
                         # Check to remove duplicate intersections
                         if nodes[node] not in POIs:
                             POIs.append(nodes[node])
+                            if nodes[node]["id"] not in nodes_ids:
+                                nodes_ids.append(nodes[node]["id"])
+                                unique_POIs.append(nodes[node])
+
     if amenity is not None and len(amenity) != 0:
         for objs in range(len(amenity)):
             POIs.append(amenity[objs])
-    return POIs  # POIs is a list of all points of interest
+            unique_POIs.append(amenity[objs])
+            # POIs is a list of all points of interest
+            # unique_POIs is an unduplicated list of all points of interest
+    return POIs, unique_POIs
 
 
 def OSM_preprocessor(processed_OSM_data, POIs, amenity):
--- a/preprocessors/openstreetmap/osm_service.py
+++ b/preprocessors/openstreetmap/osm_service.py
@@ -27,6 +27,11 @@ def create_bbox_coordinates(distance, lat, lon):
     """ Convert lat/lon from radians back to degrees """
     lat_min, lon_min = degrees(lat_min), degrees(lon_min)
     lat_max, lon_max = degrees(lat_max), degrees(lon_max)
+    """ Convert lat/lon to 7 decimal places """
+    lat_min = round(lat_min, 7)
+    lat_max = round(lat_max, 7)
+    lon_min = round(lon_min, 7)
+    lon_max = round(lon_max, 7)
     bbox_coordinates = [lat_min, lon_min, lat_max, lon_max]
     return bbox_coordinates
 
@@ -96,20 +101,27 @@ def get_timestamp():
     return timestamp
 
 
-def process_streets_data(OSM_data):
+def process_streets_data(OSM_data, bbox_coordinates):
     """Retrieve inteterested street information from the requested OSM data"""
     try:
         processed_OSM_data = []
+        lat_min = bbox_coordinates[0]
+        lat_max = bbox_coordinates[2]
+        lon_min = bbox_coordinates[1]
+        lon_max = bbox_coordinates[3]
         for way in OSM_data.ways:
             node_list = []
             for node in way.nodes:
-                node_object = {
-                    "id": int(node.id),
-                    "lat": float(node.lat),
-                    "lon": float(node.lon),
-                }
-                if node_object not in node_list:
-                    node_list.append(node_object)
+                # Extract only nodes within the boundary
+                if node.lat >= lat_min and node.lat <= lat_max:
+                    if node.lon >= lon_min and node.lon <= lon_max:
+                        node_object = {
+                            "id": int(node.id),
+                            "lat": float(node.lat),
+                            "lon": float(node.lon),
+                        }
+                        if node_object not in node_list:
+                            node_list.append(node_object)
             # Convert lanes to integer if its value is not None
             lanes = way.tags.get("lanes")
             if lanes is not None:
@@ -297,6 +309,10 @@ def allot_intersection(processed_OSM_data, inters_rec_up
 def get_amenities(bbox_coord):
     # Send request to OSM to get amenities which are part of
     # points of interest (POIs)
+    lat_min = bbox_coord[0]
+    lat_max = bbox_coord[2]
+    lon_min = bbox_coord[1]
+    lon_max = bbox_coord[3]
     try:
         amenities = server_config2(defaultServer, bbox_coord)
     except Exception:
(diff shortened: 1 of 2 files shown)


samueljohnsegun148:
Thank you @JRegimbal. The issues of duplicate POIs/intersections and the return of nodes outside the expected range raised in #507 and #510 respectively are now completely being addressed with this PR. I've assigned it to you for a further review.

preprocessors/openstreetmap/osm_service.py
```diff
@@ -27,6 +27,11 @@ def create_bbox_coordinates(distance, lat, lon):
     """ Convert lat/lon from radians back to degrees """
     lat_min, lon_min = degrees(lat_min), degrees(lon_min)
     lat_max, lon_max = degrees(lat_max), degrees(lon_max)
+    """ Convert lat/lon to 7 decimal places """
```
JRegimbal, line 30: Could you explain why it is necessary to round here?

preprocessors/openstreetmap/osm_service.py
```diff
@@ -96,49 +101,60 @@ def get_timestamp():
...
+                            node_list.append(node_object)
+            # Check if the "node_list" for a way is not empty.
+            # Otherwise all its nodes are outside the boundary, so exclude the
+            # way.
+            if node_list:
+                # Convert lanes to integer if its value is not None
+                lanes = way.tags.get("lanes")
+                if lanes is not None:
+                    lanes = int(lanes)
+                else:
```
JRegimbal, line 133: Why are these `else` statements included? They don't seem to do anything.

preprocessors/openstreetmap/osm_service.py
```diff
@@ -314,79 +334,93 @@ def get_amenities(bbox_coord):
...
-                                "name") and value != node.tags.get("amenity"):
-                            if key not in amenity_record:
-                                amenity_record[key] = value
-
-                # Delete keys with no value
-                amenity_record = dict(
-                    x for x in amenity_record.items() if all(x))
-                amenity.append(amenity_record)
+                # Extract only amenities(under nodes) within the boundary
+                if node.lat >= lat_min and node.lat <= lat_max:
```
JRegimbal, line 343: Recommend including lat and lon checking as one `if` statement. There isn't a case where we would care about being in bounds for one and not the other and this makes indentation problems more likely during future maintenance.

preprocessors/openstreetmap/osm_service.py
```diff
@@ -314,79 +334,93 @@ def get_amenities(bbox_coord):
...
-                        if value != way.tags.get(
-                                "name") and value != way.tags.get("amenity"):
-                            if key not in amenity_record:
-                                amenity_record[key] = value
-                # Delete keys with no value
-                amenity_record = dict(
-                    x for x in amenity_record.items() if all(x))
-                amenity.append(amenity_record)
+                # Extract only amenities(under ways) within the boundary
+                if way.center_lat >= lat_min and way.center_lat <= lat_max:
```
JRegimbal, line 370: Could you include in your comment what `center_lat` and `center_lon` actually indicate here? If it's the center of an entire way, I'm not sure this is correct.

preprocessors/openstreetmap/osm_service.py
```diff
@@ -398,10 +432,17 @@ def enlist_POIs(processed_OSM_data1, amenity):
...
+                                unique_POIs.append(nodes[node])
+
     if amenity is not None and len(amenity) != 0:
         for objs in range(len(amenity)):
             POIs.append(amenity[objs])
-    return POIs  # POIs is a list of all points of interest
+            unique_POIs.append(amenity[objs])
+            # POIs is a list of all points of interest
+            # unique_POIs is an unduplicated list of all points of interest
+    return POIs, unique_POIs
```
JRegimbal, line 445: I don't see why it's necessary to return both a list with duplicates and a deduplicated list. Could you please explain this?

JRegimbal (changes requested):

preprocessors/openstreetmap/osm_service.py
samueljohnsegun148: The approximation to seven decimal places does not in any way introduce an additional effect. It's just for uniformity’s sake since every other float variable (i.e., lat/lon) is represented in 7d.p.  So, I feel it looks better than having long float numbers as in ```"bounds": {
			"latitude": {
				"min": 49.89653157839408,
				"max": 49.898330221605924
			},
			"longitude": {
				"min": -97.20479051975164,
				"max": -97.20199828024836 ```

preprocessors/openstreetmap/osm_service.py
samueljohnsegun148:
```<way id="119842495">
    <nd ref="1345230863"/>
    <nd ref="216192133"/>
    <nd ref="1283567868"/>
    <tag k="highway" v="secondary"/>
    <tag k="lanes" v="2"/>
    <tag k="maxspeed" v="50"/>
    <tag k="name" v="Rue du Pont"/>
    <tag k="surface" v="asphalt"/>
  </way>```
Way keys such as the lanes, oneway, etc. usually have their raw values in strings. Meanwhile, the lanes are meant to be in integers while the other in Boolean. The if-statement does the conversion. There are also very rare cases where the "lanes" or "oneway" key has no value (e.g., ```"lanes": " "```). Any attempt to make a conversion in that case could produce a Python "ValueError". So, the idea in such a case is to accept the "empty string" as it is, which is what the else-statement is doing. The key would eventually be removed in line 156, since it has no value.

preprocessors/openstreetmap/osm_service.py
samueljohnsegun148: Points of interest (e.g. amenities) can be found under three categories in OSM: nodes, ways and relations. So, a point of interest (e.g. a restaurant) may appear as a single point (i.e. as a node). However, there could also be a situation where different restaurants are combined to be one entity (i.e. a combination of nodes forming a way, or even a combination of ways for a relation). In that case, an overpass API uses "center_lat/center_lon" to determine the latitude and the longitude of the center of such a way or relation, https://python-overpy.readthedocs.io/en/latest/api.html#overpy.Relation.center_lat, So, in order to ensure that all the POIs in a given area are captured, one has to check those in the nodes, ways and relations, and get them extracted altogether. Kindly also note that the term "way" does not necessarily have to be street in OSM, building, structures, etc. can equally be categorized as ways (closed ways) in OSM, depending on the context of application.

https://wiki.openstreetmap.org/wiki/Elements#:~:text=Elements%20are%20of%20three%20types,how%20other%20elements%20work%20together)., https://wiki.openstreetmap.org/wiki/Node, https://wiki.openstreetmap.org/wiki/Way#:~:text=An%20OSM%20way%20is%20specified,figure%20of%20the%20Earth%20used, https://wiki.openstreetmap.org/wiki/Relation

preprocessors/openstreetmap/osm_service.py
samueljohnsegun148: I could observe some level of redundancy in the code actually.  I was only initially trying to avoid the possibility of making multiple alterations to the code, but I've also now realized that it doesn't have to be. So, I've corrected it now.

preprocessors/openstreetmap/osm_service.py
samueljohnsegun148: Changes effected as recommended.

samueljohnsegun148:
I'm sorry for not assigning it to you. I just realized that I was doing it the wrong way.

preprocessors/openstreetmap/osm_service.py
JRegimbal: Look doesn't terribly bother me since this isn't meant for people to directly work with outside of prototyping and debugging. It also probably isn't enough to lead to any significant changes, so it's fine staying. If there were a reason for it other than aesthetics, it would just need to be documented.

preprocessors/openstreetmap/osm_service.py
JRegimbal: I still do not see what you are accomplishing. You're setting the variable to itself which is `None`, would it not be filtered out anyway without these extra lines?

JRegimbal (commented):
One last bit of clarification, but nearly ready to be tested and merged.

samueljohnsegun148:
You're right! The else-statement is redundant. Changes effected. The new approximation of lat/lon to 7d.p. is now equally undone.

JRegimbal (approved):
Briefly testing on Unicorn.

JRegimbal:
No obvious bugs, merging.

PR #525, opened 2022-11-16
Docker debugging

namdar-nejad:

## Required Information

## Pull Request Checklist

- [ ] The Docker image builds


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "docker/schemas"]
 	path = schemas
 	url = git@github.com:Shared-Reality-Lab/IMAGE-server.git
-	branch = schemas
+	branch = ner-image
```
JRegimbal, line 4: Should not be changing submodule branch! Merge should be made from ner-image into schemas *before* this gets merged!

docker-compose.yml
```diff
@@ -38,6 +38,13 @@ services:
         labels:
             ca.mcgill.a11y.image.preprocessor: 2
             ca.mcgill.a11y.image.port: 5000
+    ner:
+        image: ner:latest
```
JRegimbal, line 42: Would not be this image tag, but I would need to create the CI/CD workflow file for this.

docker.override.yml
```diff
@@ -0,0 +1,9 @@
+version: "3.9"
```
JRegimbal, line 1: Override should NOT be committed

docker-compose.yml
```diff
@@ -38,6 +38,13 @@ services:
...
+        env_file:
+            - ./config/ner.env
```
JRegimbal, line 44: `ner.env` is an empty file. Should be deleted, should not be loaded here.

preprocessors/ner/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ubuntu:20.04
+
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
```
JRegimbal, line 3: Do you specifically need JRE 8 for this to work??

preprocessors/ner/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN apt-get update -y \
+        && apt-get install -y software-properties-common \
+        && add-apt-repository ppa:deadsnakes/ppa \
+        && apt-get install openjdk-8-jdk -y \
+        && apt-get install python3-pip -y \
+        && export JAVA_HOME \
+        && apt-get clean \
+        && rm -rf /var/lib/apt/lists/* \
+        && apt install python3-pip
```
JRegimbal, line 13: This should come before clean and `rm -rf /var/lib/apt/lists/*`.

preprocessors/ner/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ubuntu:20.04
+
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
+
+RUN apt-get update -y \
+        && apt-get install -y software-properties-common \
+        && add-apt-repository ppa:deadsnakes/ppa \
```
JRegimbal, line 7: What is this repository used to get?

preprocessors/ner/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+        && apt install python3-pip
+
+WORKDIR /app
+RUN adduser --disabled-password python
+ENV PATH="/home/python/.local/bin:${PATH}"
+
+COPY /schemas /app/schemas
+COPY /preprocessors/ner/ /app
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y python3-opencv gcc wget git && rm -rf /var/lib/apt/lists/*
```
JRegimbal, line 22: To confirm, are you currently using opencv/gcc/wget? IIRC you *are* using git.

preprocessors/ner/clipscore.py
```diff
@@ -0,0 +1,343 @@
+'''
```
JRegimbal, line 1: For my own note, is this entirely taken from CLIPscore directly?

preprocessors/ner/ner.py
```diff
@@ -0,0 +1,219 @@
...
+    # calculate the clipscore
+    score = round(clipscore.main(parameters)['1']['CLIPScore'], 3)
+    # compute the NERs
+    ners = stanford_ner(text)
+    ner_data = []
+    for i in ners:
+        my_dict = {}
+        my_dict['value'] = i[0]
+        my_dict['tag'] = i[1]
+        my_dict['index'] = find_index(text, i[0])
```
JRegimbal, line 183: Worried about the efficiency of this `find_index` function being called for each item rather than iterating once...not blocking for now, just noting.

preprocessors/ner/ner.py
```diff
@@ -0,0 +1,219 @@
...
+        logging.error(e)
+        return jsonify("Invalid Preprocessor JSON format"), 400
+
+    # ------ START COMPUTATION ------ #
+
+    name_ = "1"
+    captions = {}                            # dict to store the captions
+    text = get_alt(content)                  # extract alt text and save it in the caption dict
+    captions[name_] = text
+    html_ = content['graphic'] # get the image html content
```
JRegimbal, line 159: Note: you are *not* guaranteed to get a `graphic` key! Many IMAGE requests do not have a blob (embedded maps and embedded charts) and will throw an error here. Please check if the graphic tag exists, return a 204 code if it does not, and then continue on as normal.

preprocessors/ner/ner.py
```diff
@@ -0,0 +1,219 @@
...
+    
+    # save pic and caption to directory for clipscore evaluation
+    with open(CONTEXT_DIR+'/captions.json', "w") as outfile:
```
JRegimbal, line 162: Note that if gunicorn is running multiple workers, they will all write to this same "captions.json" file. This should have a randomized name or something that will be unique...

preprocessors/ner/ner.py
```diff
@@ -0,0 +1,219 @@
...
+    name = "ca.mcgill.a11y.image.preprocessor.ner"
+
+    response = {
+        'request_uuid': request_uuid,
+        'timestamp': int(timestamp),
+        'name': name,
+        'data': data
+    }
+
+    try:
```
JRegimbal, line 205: Good! Will eventually be necessary to check against the data schema for NER (that you are still working on).

JRegimbal (changes requested):
Some changes in here, though looking quite good!

Question though: Is it not possible to download the jar and gz files as part of the build process? Having them committed into git is not ideal. This is not blocking in any case though.

preprocessors/ner/clipscore.py
namdar-nejad: The code is from clipscore, but I had to make some changes to:
1. combine it into one single python script
2. make it usable within other python scripts, rather than through the command line.

preprocessors/ner/ner.py
namdar-nejad: I rewrote it in the new code, the new approach also solves the problem with had with repeated tokens.

preprocessors/ner/ner.py
namdar-nejad: CONTEXT_DIR = tempfile.mkdtemp()
IMAGE_DIR = tempfile.mkdtemp()

I believe tempfile creates randomized names for the dirs: https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp

But just to be safe, I add an extra "ner_"+random(0, 999) prefix to the dir names

preprocessors/ner/ner.py
namdar-nejad: Not sure I understand this, I thought I'm doing it just below it:

```
    try:
        validator = jsonschema.Draft7Validator(schema, resolver=resolver)
        validator.validate(response)
    except jsonschema.exceptions.ValidationError as e:
        logging.error(e)

    logging.debug("Sending response")
    return response
```

preprocessors/ner/Dockerfile
namdar-nejad: This is mentioned on the Stanford ner website:
`Either make sure you have or get [Java 8](http://java.com/) or consider running an earlier version of the software (versions through 3.4.1 support Java 6 and 7).. (https://nlp.stanford.edu/software/CRF-NER.html#Download)`

For what I understand version 8 is the latest that would work. right?

preprocessors/ner/Dockerfile
namdar-nejad: I'm using `FROM ubuntu:20.04` as my base container since it has java pre-installed. I'm using the repos to install python. Not sure it there's a better way.

preprocessors/ner/Dockerfile
namdar-nejad: I'm using gcc and git, not opencv and get. I removed those.

docker-compose.yml
namdar-nejad: Should I change it to `ner:unstable` ?

preprocessors/ner/ner.py
JRegimbal: Great, missed this.

preprocessors/ner/ner.py
JRegimbal: This only checks that you're returning something in the expected response format. It doesn't check that the content of the `data` key actually makes sense for the NER preprocessor `name`.

preprocessors/ner/Dockerfile
JRegimbal: There is definitely a better way - Python is in the main ubuntu repos. You don't need to add another PPA for something like this. Python 3.x is available as the `python3` package since historically Python 2.x was available as `python`. Since 2 is past EoL, the `python` package no longer exists. In fact...you're probably installing python3 as a dependency of `python3-pip`. I'm not sure why you're installing it twice though (lines 9 and 11).

JRegimbal (commented):
Looking good! Some minor issues with the Dockerfile, but really this is probably good to go as is when I add the workflow file and when the schemas PR is ready.

JRegimbal:
Actually, I've added a workflow file for this preprocessor. It should start running automatically once you resolve the merge conflicts in `docker-compose.yml` and `build.yml`.

docker-compose.yml
```diff
@@ -38,6 +38,11 @@ services:
         labels:
             ca.mcgill.a11y.image.preprocessor: 2
             ca.mcgill.a11y.image.port: 5000
+    ner:
+        image: ner:latest
```
JRegimbal, line 42:
```suggestion
        image: ghcr.io/shared-reality-lab/image-preprocessor-ner:unstable
```

preprocessors/ner/ner.py
```diff
@@ -0,0 +1,257 @@
+# flake8: noqa
```
JRegimbal, line 1: Why did you add `noqa` to this file? If I understand correctly, you wrote all of this?

JRegimbal (changes requested):
Just minor things at this point.

preprocessors/ner/ner.py
```diff
@@ -40,48 +39,48 @@
...
+    :my_html: the html to save
+    :name: name of the to save
+    :out_dir: directory to save the image to
+    """
     path_ = os.path.abspath(f"{out_dir}")
     image_b64 = my_html.split(",")[1]
     binary = base64.b64decode(image_b64)
-    with open(f"{path_}/{name}.png","wb") as file:
+    with open(f"{path_}/{name}.png", "wb") as file:
         file.write(eval(str(binary)))
```
JRegimbal, line 58: Just noticed this...with the binary data coming from user input, calling `eval` basically just opens us to RCE. A quick local test shows the following works:
```suggestion
        file.write(binary)
```
On another note, is CLIPscore fine loading from a .png extension even if the encoding is different (e.g., JPEG)?

JRegimbal (changes requested):
Basically good, just caught one potential issue with the code. Ready to merge once addressed.

preprocessors/ner/ner.py
namdar-nejad: I'm not sure about the encoding, the documentation, or code, don't mention anything about the encoding.

JRegimbal (approved):
Looks good!

Issue #543, opened 2023-01-18
Semantic segmentation timing out on all inputs


JRegimbal:
The semantic segmentation preprocessor is regularly not responding and timing out to inputs in production. Logs for each request follow what is below (timestamps given for clarity)
```
image-semantic-segmentation  | 2023-01-18T20:14:20.144991143Z [2023-01-18 20:14:20 +0000] [807] [DEBUG] POST /preprocessor
image-semantic-segmentation  | 2023-01-18T20:14:20.145278606Z DEBUG:root:Received request
image-semantic-segmentation  | 2023-01-18T20:14:20.165502569Z INFO:root:Schemas loaded
image-semantic-segmentation  | 2023-01-18T20:14:50.974534070Z [2023-01-18 20:14:50 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:807)
image-semantic-segmentation  | 2023-01-18T20:14:51.999076100Z [2023-01-18 20:14:51 +0000] [1] [WARNING] Worker with pid 807 was terminated due to signal 9
image-semantic-segmentation  | 2023-01-18T20:14:52.000474719Z [2023-01-18 20:14:52 +0000] [885] [INFO] Booting worker with pid: 885
```
Note that this is not the semantic segmentation process returning after 30 seconds, but gunicorn itself killing the process and restarting. When requests are received, there is no significant increase in resource usage - it appears that the process is hanging.

This will not always occur. It is unclear what conditions lead to this state at this time.

jeffbl:
@JRegimbal Just to be sure, can you include link to a photo that definitely reproduced the issue?

JRegimbal:
On the IMAGE homepage, https://image.a11y.mcgill.ca/pages/imgs/pexels-photo-5926382.jpeg and https://image.a11y.mcgill.ca/images/pexels-photo-3584998_mod.jpeg. Note that I've restarted the container already and these work, so this seems to be more complicated than "these photos do not work".

jeffbl:
As discussed, @rohanakut please add sufficient debug outpu to the semseg code to be able to capture what is going on when this happens again.

rohanakut:
@jeffbl I have opened a PR for the same

jeffbl:
Happening right now, on multiple photos I've tried. This is log from the bear picture on this page:
https://image.a11y.mcgill.ca/pages/tutorial.html

```
[...]

PR #557, opened 2023-02-05
Added debug tags in semantic segmentation

rohanakut:
This PR is wrt to #543 . Currently the issue is not reproducible. Hence I have added a few debug tags for future reproducibility of this issue. the code has been tested on unicorn. The following was the output

Screenshot 2023-02-05 at 12 07 47 PM

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

### DIFF
--- a/preprocessors/semanticSeg/segment.py
+++ b/preprocessors/semanticSeg/segment.py
@@ -57,6 +57,7 @@ def visualize_result(img, pred, index=None):
     if index is not None:
         pred = pred.copy()
         pred[pred != index] = -1
+    logging.info("encoding detected segmets with unique colors")
     pred_color = colorEncode(pred, colors).astype(numpy.uint8)
     nameofobj = names[index + 1]
     return pred_color, nameofobj
@@ -72,8 +73,10 @@ def findContour(pred_color, width, height):
     ret, thresh = cv2.threshold(gray_image, 10, 255, cv2.THRESH_BINARY)
     contours, hierarchy = cv2.findContours(
         thresh, cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_NONE)
+    logging.info("drawing all contours")
     cv2.drawContours(image, contours, -1, (0, 255, 0), 2)
     # removes the remaining part of image and keeps the contours of segments
+    logging.info("deleting remainder of the image except the contours")
     image = image - dummy
     centres = []
     area = []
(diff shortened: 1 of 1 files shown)


preprocessors/semanticSeg/segment.py
```diff
@@ -72,15 +73,18 @@ def findContour(pred_color, width, height):
...
     # removes the remaining part of image and keeps the contours of segments
+    logging.info("deleting remainder of the image except the contours")
     image = image - dummy
     centres = []
     area = []
     totArea = 0
     send_contour = []
     flag = False
     # calculate the centre and area of individual contours
+    logging.info("computing individual contour metrics")
```
jeffbl, line 87: suggest including len(contours) here so we can see if there are an outrageous number of iterations here

preprocessors/semanticSeg/segment.py
```diff
@@ -106,12 +110,14 @@ def findContour(pred_color, width, height):
             contour_indi[j][1] = float(float(contour_indi[j][1]) / height)
```
jeffbl, line 110: likewise, would log the number of expected iterations of this loop before entering

preprocessors/semanticSeg/segment.py
```diff
@@ -143,6 +150,7 @@ def run_segmentation(url,
     scale_size = np.float(1500.0 / np.float(max(height, width)))
     # scale down an image to avoid OOM error
     if scale_size <= 1.0:
+        logging.info("scaling down an image")
```
jeffbl, line 153: please add original and new size logging

preprocessors/semanticSeg/segment.py
```diff
@@ -162,6 +170,7 @@ def run_segmentation(url,
     output_size = img_data.shape[1:]
     with torch.no_grad():
         # get segmentation results
+        logging.info("running segmentation model")
```
jeffbl, line 173: If the below lines may take substantial time (especially in degenerate cases), suggest having logging for each of the steps so we know where it halts and dies. Also, in line 181, there is another loop where the number of iterations should ideally be logged so we can see if anything crazy is happening.

jeffbl (commented):
Great to have more logging, but I'm not sure what is here will fully help us narrow down the problem when it dies again. Any step that might take significant time should have a log line, and the logging should ideally output key things like number of expected iterations, since if these numbers are crazy, it can help narrow down the issue. In general, anytime we can be more specific about what is happening, the better.

PR #561, opened 2023-02-07
Added additional debug tags

rohanakut:
This PR is wrt to #543. I have added additional debug tags and start and end time for every loop to check where the timeout occurs.

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

* [ ] I followed applicable coding standards (e.g., [PEP8](https://pep8.org/))

### DIFF
--- a/preprocessors/semanticSeg/segment.py
+++ b/preprocessors/semanticSeg/segment.py
@@ -73,7 +73,7 @@ def findContour(pred_color, width, height):
     ret, thresh = cv2.threshold(gray_image, 10, 255, cv2.THRESH_BINARY)
     contours, hierarchy = cv2.findContours(
         thresh, cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_NONE)
-    logging.info("drawing all contours")
+    logging.info("Total contours detected are: {}".format(len(contours)))
     cv2.drawContours(image, contours, -1, (0, 255, 0), 2)
     # removes the remaining part of image and keeps the contours of segments
     logging.info("deleting remainder of the image except the contours")
@@ -105,9 +105,11 @@ def findContour(pred_color, width, height):
         centre_down = [centre_indi[0] / width, centre_indi[1] / height]
         area_down = area_indi / (width * height)
         contour_indi = contour_indi.tolist()
+        logging.info("Iterating through individual contour points with starting time: {}".format(str(time.time())))
         for j in range(len(contour_indi)):
             contour_indi[j][0] = float(float(contour_indi[j][0]) / width)
             contour_indi[j][1] = float(float(contour_indi[j][1]) / height)
+        logging.info("Iteration completed with time: {}".format(str(time.time())))
         send_contour.append({"coordinates": contour_indi,
                             "centroid": centre_down, "area": area_down})
     logging.info("computed all metrics!!")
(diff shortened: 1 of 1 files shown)


preprocessors/semanticSeg/segment.py
```diff
@@ -73,7 +73,7 @@ def findContour(pred_color, width, height):
     ret, thresh = cv2.threshold(gray_image, 10, 255, cv2.THRESH_BINARY)
     contours, hierarchy = cv2.findContours(
```
jeffbl, line 79: What are the failure cases for this? Should this be in a try/catch? If this throws an error, what happens? In general, there seems to be minimal error checking, so if opencv or any of the pytorch calls throw an error, what happens?

preprocessors/semanticSeg/segment.py
```diff
@@ -149,12 +151,14 @@ def run_segmentation(url,
...
     scale_size = np.float(1500.0 / np.float(max(height, width)))
     # scale down an image to avoid OOM error
+    logging.info("original size of image: {}".format(pil_image.shape ))
     if scale_size <= 1.0:
         logging.info("scaling down an image")
         height = np.int(height * scale_size)
         width = np.int(width * scale_size)
         pil_image = cv2.resize(pil_image, (width, height),
                                interpolation=cv2.INTER_AREA)
+        logging.info("scaled down image size is: {}".format(pil_image.shape))
```
jeffbl: Suggest (using recommended wording "graphic"):
`graphic original dimensions `
`graphic scaled dimensions `

preprocessors/semanticSeg/segment.py
```diff
@@ -105,9 +105,11 @@ def findContour(pred_color, width, height):
         centre_down = [centre_indi[0] / width, centre_indi[1] / height]
         area_down = area_indi / (width * height)
         contour_indi = contour_indi.tolist()
+        logging.info("Iterating through individual contour points with starting time: {}".format(str(time.time())))
         for j in range(len(contour_indi)):
             contour_indi[j][0] = float(float(contour_indi[j][0]) / width)
             contour_indi[j][1] = float(float(contour_indi[j][1]) / height)
+        logging.info("Iteration completed with time: {}".format(str(time.time())))
```
jeffbl: Suggest `End contour iteration` so that it is clear what is ending. (see comment below about making start/end wording consistent)

jeffbl (changes requested):
I'd been focusing on the logging, but is there other error checking that should be done in this file? Also, see comment about timestamps, since I'm not sure they are necessary.

preprocessors/semanticSeg/segment.py
rohanakut: @jeffbl if there are no contours in the image then the "contours" variable will be a empty list. In that case the for loop would not be iterated and we would simple get an empty list. I don't think there is a need for try catch as the code internally handles that use case.

preprocessors/semanticSeg/segment.py
```diff
@@ -202,13 +196,11 @@ def run_segmentation(url,
...
         logging.info(
-            "running segmentation model"
-            " with start time: {}".format(str(time.time())))
+            "running segmentation model")
         scores = segmentation_module(singleton_batch,
                                      segSize=output_size)
         logging.info(
-            "run successful with"
-            " end time: {}".format(str(time.time())))
+            "run successful with")
```
jeffbl, line 203: since you removed the timestamp, this no longer makes sense. suggest "finished segmentation: success" so that it is clear what is finishing

preprocessors/semanticSeg/segment.py
jeffbl: ok. I'm trying to make sure that if there is a failure in any of these calls, that we'll get a meaningful message that will help us track it down. e.g., if it runs out of memory, if the input data is bad somehow, etc. If you're confident that an error will be logged if something goes wrong here, then all good.

jeffbl (approved):

jeffbl:
Just a couple other comments, then should go to @jaydeepsingh25 for final approval. @jaydeepsingh25 the goal is that when this fails on pegasus again, we have enough debug information to track down where the problem is, so if you have other recommendations, please include in your review. thanks!

Issue #603, opened 2023-03-05
As an IMAGE user who is blind, I want to experience more accurate and useful photo breakdowns into regions of interest, so that I can get a better understanding of photos beyond individual objects detected.


rbazin:

rbazin:
- [ ] #577 
- [ ] #578 
- [ ] Deploy model as default on test and production

Issue #572, opened 2023-02-13
STORY : As a preprocessor developer, I want to compare MSEG and other SemSeg candidates performance in a real-world scenario to the current implementation, so that I can potentially improve semseg used in IMAGE.


rbazin:
- [x] #574
- [x] #575
- [x] #576

rbazin:
In my opinion, it’s worth replacing the current standalone model (Semseg) by a newer semantic segmentation framework, because they bring modularity, which implies flexibility, support over time, ease of use and scalability.

My framework of choice would be MMsegmentation, which is open source, has a permissive license, exposes many models with pretrained weights (trained on ADE20K), has a modular architecture, is easy to use, has good documentation, and shows promising speed and accuracy.

You can find the whole report [here](https://docs.google.com/document/d/1O_o4Mqz7cMsNRZ0kbz2dGhm3sz8oEsVONsmaL-9g6Hc/edit?usp=sharing).

PR #639, opened 2023-03-29
Semseg preprocessor replacement

rbazin:
In this PR, I present a **new semantic segmentation preprocessor** that I created for #603  . It relies on [MMSegmentation](https://github.com/open-mmlab/mmsegmentation) framework. The main benefits of this preprocessor are the improved image segmentations and the increased modularity for possible future improvements.

I added the full frameworks review from #572 in the "docs" folder.

## Tests

The whole preprocessor has been tested on unicorn, over the whole IMAGE-graphics-test set. The preprocessor uses and validates the **exact same schemas** as the previous semantic segmentation preprocessor.

Here are examples of the segmentation, taken from the extension, connected to unicorn :
1. Input :

boris

Output : 
Screenshot 2023-03-24 at 4 17 32 PM

Classes found : Person, wall, flag

2. Input :

castle

Output : 
Screenshot 2023-03-24 at 4 14 56 PM

Classes found : building, sky, tree, water

## Ressources used 

All tests were conducted using the NVIDIA Titan XP GPU.  

Using nvidia-smi and top commands, I monitored both the GPU and CPU loads.

- The average CPU load during the segmentations is between 300% and 400%.
- Most of the time, the GPU load was at 3GB, but sometimes it toped at around 11GB.

**Note** :  I encountered one unusual error during the testing phase on the IMAGE-graphics-test set. The error was that the segmentation failed for the image called "space.jpg", which has a resolution of 3x15 (so only a couple of pixels). For this reason, I reckon that the image shouldn't have been part of the test set in the first place (it's not representative of anything on the internet), and that no fix is needed for the preprocessor to handle this image.

---

## Required Information

## Coding/Commit Requirements


rbazin:
Fixed the things you asked me to review

jeffbl:
@jaydeepsingh25 yes, I am trying to push people to put ongoing documentation into git directly. The photos shouldn't bloat up the repo too much, and we don't have a good way of having external resources linked. Let me know if you strongly disagree, but we'll likely do more of this (trying to avoid massive churn). Obviously, if we need to do videos or large graphic additions, we'll need to do that outside of the main repo.

jeffbl:
@rbazin we purposely put in some very small graphics and other boundary cases into the test set to make sure nothing falls completely apart on such inputs. I don't think we strip out small graphics like this, so they could currently come from the browser extension. So the question is whether the preprocessor breaks in a bad way, or returns an appropriate error code that it can't produce results. It should do the latter, and continue working on future inputs (i.e., no crashing, etc)

rbazin:
About the very small image, the error is caught and I return a 500 error code. Also, I updated the code so that all segmentation errors are caught and return a 500 error code. I believe this way the segmentation preprocessor shouldn't break the pipeline. If everything is fine on your side, we can move with the merge.

Issue #662, opened 2023-05-05
STORY: As a handler developer, I want a service that translates English text into French, so that I can use it to generate French language renderings.


notkaramel:
- [ ] Search for alternative translation API other than DeepL
- [ ] Create a new service using DeepL API
- [ ] Setup Docker container for the service
- [ ] Test/Validate the service by sending a sample inputs in English and evaluate
- [ ] Deploy language translation service on unicorn for testing

notkaramel:
Here is a list of all Translation APIs I've found so far (as of May 9th): [ListOfAPITranslator - Google Sheet](https://docs.google.com/spreadsheets/d/1L3N3hPrB77bh-o3SpnUPLhmNl5A1YA9cOgvCjlc0uZ4/edit?usp=sharing)

Newly added HuggingFace model to the list & the new tasks for the STORY were the result of a discovery on May 11th, 2023. More updates on benchmark comparison between models and implementation will be reported here soon!

notkaramel:
Schema for request validation has been added: #687 

As of May 19th, we decided to go with [Helsinki-NLP/opus-mt-en-fr](https://huggingface.co/Helsinki-NLP/opus-mt-en-fr) model for French translation. It provides the most accurate result with both short (1 word) and long (sentence) queries within the least amount of time (in comparison to [t5-small](https://huggingface.co/t5-small) and [google/flan-t5-small](https://huggingface.co/google/flan-t5-small).

More information is provided in the [README.md](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/multilang-service/services/multilang-support/README.md) file of the repo.

notkaramel:
New tasks added (May20), focusing on schema validation, `Gunicorn` implementation, exception handling (GPU selection, empty segment(s), etc.), softcode to add more languages, attribution to the LLM source code, and testings on `unicorn` server.

notkaramel:
Note that to optimise model's generation speed, we can simply add/modify parameters for:
```py
def generate_new_tensor(input_ids:torch.Tensor) -> torch.Tensor:
    '''
    STEP 2: Translate the input_ids tensor to an output query.
    [...]
    '''
    return MODEL.generate(input_ids, \
        max_new_tokens=MAX_NEW_TOKENS, \
        max_time=MAX_TIME, \
        #... 
        )
```

notkaramel:
[...]

PR #688, opened 2023-05-20
Translation Service Implementation 

notkaramel:
# PR for STORY #662.

## Major changes include:
- Addition of `multilang-support` to `services`
- Implementation of a translation service that receives HTTP request from `photo-audio-handler`
- Using pre-trained LLMs to translate from English to French (and other languages in the future)

## Test format
- An HTTP request send a JSON message containing the following fields (following the translation schema):
  - "segments": an array of segments to be translated
  - "src_lang": source language (default: 'en'). This field is optional
  - "tgt_lang": target language. This field is required
- Each test records:
  - HTTP Response Status Code 
  - Translation time taken (in seconds)
  - The translated segments (this requires developer's or user's evaluation on accuracy)
- The JSON file is sent to the service via `curl` POST request

## Test performed:
- [x] No source language, default to English as source
- [x] Invalid Requests:
  - [x] No target language
  - [x] No segments
- [x] Empty segments
- [x] Target language other than French, must return `Not Implemented`
- [x] Same language (returns nothing, but should be returning a message)

> Please ensure you've followed the checklist and provide all the required information *before* requesting a review.

> Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

### DIFF
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -2,8 +2,7 @@ FROM pytorch/pytorch:latest
 
 # See README.md#docker-image for visualisation of the directory structure
 WORKDIR /app
-
-RUN apt-get update && apt-get install python3 python3-pip -y
+RUN apt-get update && apt-get install python3 python3-pip git -y
 
 # Copy requirements 
 COPY services/multilang-support/requirements.txt /app/
@@ -11,6 +10,10 @@ COPY services/multilang-support/requirements.txt /app/
 # Upgrade pip and install requirements
 RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
+
+# Clone the model
+RUN git lfs install
+RUN git clone https://huggingface.co/Helsinki-NLP/opus-mt-en-fr ./model/opus-mt-en-fr
 
 COPY schemas/services/translation.schema.json /app/
 COPY services/multilang-support/src/*.py /app/src/
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -1,0 +1,16 @@
+FROM python:3-slim
+
+# Set working directory for the container
+WORKDIR /usr/src/app
+
+COPY services/multilang-support/requirements.txt ./
+RUN apt-get update && apt-get install python3 python3-pip -y
+RUN pip3 install --upgrade pip
+RUN pip install -r requirements.txt
+
+COPY schemas ./schemas_image
+COPY services/multilang-support/app.py ./
+
+EXPOSE 5000
+ENV FLASK_APP=app.py
+CMD [ "python3", "app.py" ]
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -1,9 +1,9 @@
-FROM pytorch/pytorch:2.0.1-cuda11.7-cudnn8-runtime
+FROM pytorch/pytorch:1.12.1-cuda11.3-cudnn8-runtime
 
 COPY services/multilang-support/requirements.txt ./
 RUN apt-get update && apt-get install python3 python3-pip -y
-RUN pip3 install --upgrade pip
-RUN pip3 install -r requirements.txt
+RUN pip install --upgrade pip
+RUN pip install -r requirements.txt
 
 COPY schemas ./schemas
 COPY services/multilang-support/src ./src
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -1,13 +1,15 @@
-FROM pytorch/pytorch:1.12.1-cuda11.3-cudnn8-runtime
+FROM pytorch/pytorch:latest
 
 COPY services/multilang-support/requirements.txt ./
 RUN apt-get update && apt-get install python3 python3-pip -y
+
+# Upgrade pip and install requirements
 RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
 
-COPY schemas ./schemas
-COPY services/multilang-support/src ./src
+COPY schemas/services/translation.schema.json ./
+COPY services/multilang-support/src/* ./src
 
 EXPOSE 5000
 ENV FLASK_APP=src/app.py
-CMD [ "python3", "src/app.py" ]
+CMD [ "gunicorn", "src.app:app" ]
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -1,15 +1,20 @@
 FROM pytorch/pytorch:latest
 
-COPY services/multilang-support/requirements.txt ./
+# See README.md#docker-image for visualisation of the directory structure
+WORKDIR /app
+
 RUN apt-get update && apt-get install python3 python3-pip -y
+
+# Copy requirements 
+COPY services/multilang-support/requirements.txt /app/
 
 # Upgrade pip and install requirements
 RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
 
-COPY schemas/services/translation.schema.json ./
-COPY services/multilang-support/src/* ./src
+COPY schemas/services/translation.schema.json /app/
+COPY services/multilang-support/src/*.py /app/src/
 
 EXPOSE 5000
 ENV FLASK_APP=src/app.py
-CMD [ "gunicorn", "src.app:app" ]
+CMD [ "gunicorn", "src.translate:app", "-b", "0.0.0.0", "--log-level=debug"]
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -2,7 +2,7 @@ FROM pytorch/pytorch:latest
 
 # See README.md#docker-image for visualisation of the directory structure
 WORKDIR /app
-RUN apt-get update && apt-get install python3 python3-pip git -y
+RUN apt-get update && apt-get install python3 python3-pip git git-lfs -y
 
 # Copy requirements 
 COPY services/multilang-support/requirements.txt /app/
@@ -12,7 +12,8 @@ RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
 
 # Clone the model
-RUN git clone https://huggingface.co/Helsinki-NLP/opus-mt-en-fr ./models/opus-mt-en-fr
+RUN git lfs install
+RUN git clone https://huggingface.co/Helsinki-NLP/opus-mt-en-fr ./model/opus-mt-en-fr
 
 COPY schemas/services/translation.schema.json /app/
 COPY services/multilang-support/src/*.py /app/src/
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -1,16 +1,13 @@
 FROM python:3-slim
 
-# Set working directory for the container
-WORKDIR /usr/src/app
-
 COPY services/multilang-support/requirements.txt ./
 RUN apt-get update && apt-get install python3 python3-pip -y
 RUN pip3 install --upgrade pip
-RUN pip install -r requirements.txt
+RUN pip3 install -r requirements.txt
 
-COPY schemas ./schemas_image
-COPY services/multilang-support/app.py ./
+COPY schemas ./schemas
+COPY services/multilang-support/src ./src
 
 EXPOSE 5000
-ENV FLASK_APP=app.py
-CMD [ "python3", "app.py" ]
+ENV FLASK_APP=src/app.py
+CMD [ "python3", "src/app.py" ]
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3-slim
+FROM pytorch/pytorch:2.0.1-cuda11.7-cudnn8-runtime
 
 COPY services/multilang-support/requirements.txt ./
 RUN apt-get update && apt-get install python3 python3-pip -y
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -18,6 +18,6 @@ RUN git clone https://huggingface.co/Helsinki-NLP/opus-mt-en-fr ./models/opus-mt
 COPY schemas/services/translation.schema.json /app/
 COPY services/multilang-support/src/*.py /app/src/
 
-EXPOSE 8000
+EXPOSE 80
 ENV FLASK_APP=src/app.py
-CMD [ "gunicorn", "src.translate:app", "-b", "0.0.0.0", "--log-level=debug"]
+CMD [ "gunicorn", "src.translate:app", "-b", "0.0.0.0:80", "--log-level=debug"]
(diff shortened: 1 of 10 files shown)


notkaramel:
By adding logging to the program, it might increase the runtime of the translate service significantly. Currently testing on `unicorn` for observation.

jeffbl:
Removed @notkaramel assignment since was unclear if ready for review. (If not ready, please assign back to @notkaramel exclusively)

services/multilang-support/README.md
```diff
@@ -0,0 +1,52 @@
...
+### Docker Compose as an IMAGE service
+- [IMAGE-server/wiki on Services](https://github.com/Shared-Reality-Lab/IMAGE-server/wiki/2.-Handlers,-Preprocessors-and-Services#services)
+
+## Overview
+- In this implementation, we use [`Helsinki-NLP/opus-mt-en-fr`](https://huggingface.co/Helsinki-NLP/opus-mt-en-fr) as the model checkpoint to be used out of the box. It's licensed under CC-BY 4.0 License.
+- The model is a `MarianMTModel` from `transformers` library.
+- As of May 30th, generalization is not yet completed to support multiple languages. The service can currently only translate from English to French.
+  
+## Attribution
+- License: https://creativecommons.org/licenses/by-nc-sa/4.0/
```
JRegimbal, line 46: Clarify what is under this license.

services/multilang-support/requirements.txt
```diff
@@ -0,0 +1,6 @@
+Flask
```
JRegimbal, line 1: Please specify minimum version numbers to avoid unpleasant surprises later on.

services/multilang-support/src/utils.py
```diff
@@ -0,0 +1,218 @@
...
+@log
+def generate_output_tensor(input_ids: torch.Tensor) -> torch.Tensor:
+    """
+    STEP 2: Translate the input_ids tensor to an output query.
+    @param input_ids: The decoded tensor (type<torch.Tensor>) to be translated.
+    @return: Newly generated tensor using the model (type<torch.Tensor>).
+    TODO: Add parameters to control/optimize the translation.
+    """
+    # Parameters:
+    MAX_TIME = 0.5  # seconds
```
JRegimbal, line 127: Recommend not mixing in constants like this throughout the file.

services/multilang-support/src/utils.py
```diff
@@ -0,0 +1,218 @@
...
+    num_gpus = torch.cuda.device_count()
+    LOGGER.info(f"Instantiating: {MODEL_CHECKPOINT} tokenizer and model")
+
+    TOKENIZER = AutoTokenizer.from_pretrained(MODEL_CHECKPOINT)
+    LOGGER.debug("Tokenizer instantiated")
+
+    MODEL = AutoModelForSeq2SeqLM.from_pretrained(MODEL_CHECKPOINT)
+    LOGGER.debug("Model instantiated")
+
+    device_id = 0
```
JRegimbal, line 89: Have you tested this if you only make one GPU available (e.g., host ID 1 and not 0)? I cannot remember if IDs are remapped within the container or not - my suspicion is that they are not.

notkaramel:
@JRegimbal confirmed that PR is ready for review, I've just added all the tests I performed on the description of the PR. Things should be working smoothly now. Any improvement will be added in the future. 

@jeffbl apologies, I didn't know to unassign myself when PR is ready for review.

services/multilang-support/src/translate.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    try:
+        jsonschema.validate(instance=request, schema=TRANSLATION_SCHEMA)
+        return True
+    except jsonschema.exceptions.ValidationError as e:
+        LOGGER.error(f'Schema ValidationError: {e.message}')
+        return False
+
+
+@app.route("/service/translate/french", methods=["POST"])
```
JRegimbal, line 41: The path should not contain the target language since this is already included in the body of the request.

services/multilang-support/src/translate.py
```diff
@@ -0,0 +1,85 @@
...
+
+# Validate request
+
+
+def validate_request(request):
+    """
+    Validate incoming request:
+    - segment: the text to translate
+    - src_lang (defaulted to be 'en'): the source language
+    - tgt_lang (defaulted to be 'fr'): the target language
```
JRegimbal, line 30: French should not be a default target. This is in conflict with the schema which has no default.

services/multilang-support/src/translate.py
```diff
@@ -0,0 +1,85 @@
...
+with open("translation.schema.json", "r") as f:
+    TRANSLATION_SCHEMA = json.load(f)
+
+# Validate request
+
+
+def validate_request(request):
+    """
+    Validate incoming request:
+    - segment: the text to translate
```
JRegimbal, line 28: Schema specifies a `segments` key not `segment`.

services/multilang-support/src/translate.py
```diff
@@ -0,0 +1,85 @@
...
+    try:
+        source_lang = content["src_lang"]
+    except KeyError:
+        source_lang = "en"
+
+    target_lang = content["tgt_lang"]
+
+    # Handles source/target language
+    if target_lang == source_lang:
+        return jsonify("Source and target languages are the same"), 204
```
JRegimbal, line 68: This is fine, but make sure that you include this behavior in the README since 204 status codes are typically only used by preprocessors.

services/multilang-support/src/translate.py
```diff
@@ -0,0 +1,85 @@
...
+    if target_lang not in ["fr", "en"]:
+        return jsonify("Target language not yet implemented"), 501
+    if source_lang == 'en' and target_lang == 'fr':
+        # Translate, from list to list
+        translation, elapsed_time = translate_helsinki(segments)
+    else:
+        return jsonify("Internal Server Error"), 500
```
JRegimbal, line 75: This is a meaningless response and nothing is logged...

services/multilang-support/src/translate.py
```diff
@@ -0,0 +1,85 @@
...
+    # Prepare response
+    response = {
+        "src_lang": source_lang,
+        "tgt_lang": target_lang,
+        "elapsed_time_in_seconds": elapsed_time,
+        "translations": translation,
+    }
+    LOGGER.debug("- Response SENT! -")
```
JRegimbal, line 83: Suggestion: log elapsed time here.

JRegimbal (changes requested):

services/multilang-support/src/utils.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    Timing & logging decorator for functions.
+    """
+
+    def wrapper(*args, **kwargs):
+        start_time = time.time()
+        function_output = func(*args, **kwargs)
+        end_time = time.time()
+        elapsed_time = end_time - start_time
+        LOGGER.debug(f"\'{func.__name__}\' takes {elapsed_time:.3f} seconds")
```
JRegimbal, line 31: Considering the times shown in your demo, probably good to put this in ms.

JRegimbal (commented):
Q: when/from where do you get the weights?

services/multilang-support/src/translate.py
notkaramel: Edited to `/service/translate`

services/multilang-support/src/translate.py
notkaramel: Edited `src_lang` and `tgt_lang` description

services/multilang-support/src/translate.py
notkaramel: Corrected `segment` to `segments` (with the `s`)

services/multilang-support/src/translate.py
notkaramel: Moved `elapsed_time` from repsonse to debug log.
@JRegimbal Should I keep the source and/or target language in the response? My idea was to know and make sure what the handler sent, but I think it's a bit unnecessary.

services/multilang-support/src/translate.py
notkaramel: I changed it to
```py
    LOGGER.error("Service Error, unable to handle")
    return jsonify("Service Error"), 500
```
indicating that this is an edge case that the service or myself did not expected.

services/multilang-support/src/utils.py
notkaramel: Yes. Docker remaps the 2nd GPU (TITAN Xp on `unicorn`) to `cuda:0` and use it.

services/multilang-support/src/utils.py
notkaramel: Noted.
I originally didn't want to hardcode a constant like this, it will sometimes ruin the output result. However, without parameters like `max_time` or `max_new_tokens`, it produces a warning (below). Having large `max_new_tokens` might slows down the generation time, while low `max_time` prevents long query to be translated.
```yaml
UserWarning: Using `max_length`'s default (512) to control the generation length. This behaviour is deprecated and will be removed from the config in v5 of Transformers -- we recommend using `max_new_tokens` to control the maximum length of the generation.
```
I suggest creating a work item (issue) on what to do with this, will discuss with @rbazin on a solution and maybe adapt JAX implementation instead of current one.

JRegimbal: To clarify, I'm not opposed to constants! Just I prefer to have them all listed in a few places near the top of a file so that we can quickly make changes in the future. Digging through to find them is unpleasant, especially if you're not the one who wrote the code in the first place!

services/multilang-support/src/utils.py
JRegimbal: Awesome, thanks!

services/multilang-support/src/translate.py
JRegimbal: Do you indicate the input/output language codes somewhere else? Without these it'll be difficult to reproduce...

services/multilang-support/src/translate.py
JRegimbal: It's a little bit unnecessary, but not a problem and may eventually come in handy for debugging. If you'd prefer it, I'm fine with it being left in.

services/multilang-support/src/utils.py
JRegimbal: I just mean the weights for the model you're using. Usually there's a pretty clear line in the Dockerfile or Python code itself where these are downloaded ahead of time, and I didn't see one when I looked through.

notkaramel: In this implementation, the model is loaded when the program/module is started (see `instantiate()`), so it's not downloaded ahead of time.
I will look into that, the weights concept is new to me at the moment 🤔

JRegimbal: Ah no worries. How long does it take to download with a clean container? This is the same way the TTS is set up, but we had to find a way to predownload it to avoid a huge delay when the first request is received.

notkaramel: For fresh container, it takes around ~3s to download and get ready (excluding container build time). Sometimes it randomly takes up to ~10s to get ready, but it's very rare. 

I also noticed (and forgot to mention anywhere) that first request after initialising often takes more time to process compared to requests after that. A query similar to the one during demo takes ~1.5s to process on first req, but any similar queries after that only takes ~0.4s to 0.8s.

services/multilang-support/src/utils.py
notkaramel: Moved the constant to the top of the file. If there is new change or update to functionality, future PRs will cover them.

notkaramel:
Updates:
- Model is now able to be downloaded ahead of time to "local" Docker Container.
- Cloning model from [Helsinki-NLP/opus-mt-en-fr](https://huggingface.co/Helsinki-NLP/opus-mt-en-fr/tree/main), using `git-lfs`. Model's **weights** can also be found here in [pytorch_model.bin](https://huggingface.co/Helsinki-NLP/opus-mt-en-fr/blob/main/pytorch_model.bin) (301 MB)
- Instantiate time is ~2.4s (still consistent to before "predownload" changes
- First request still take more time (~twice amount of time) compared to later requests. 

@JRegimbal Should I let the script just send a dummy request to server when it's fully set up?

jeffbl:
@JRegimbal RE dummy request, might we want to consider something like this across the board, since I believe there are similar issues with other containers as well? i.e., such that we could eventually get to a point where the first request is pretty much guaranteed to time out? If not clear best way to do this, new work item could go to Jaydeep, since this is primarily a preprocessor issue (despite this specific example being a service).

services/multilang-support/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN pip install -r requirements.txt
+
+# Clone the model
+RUN git lfs install
+RUN git clone https://huggingface.co/Helsinki-NLP/opus-mt-en-fr ./models/opus-mt-en-fr
+
+COPY schemas/services/translation.schema.json /app/
+COPY services/multilang-support/src/*.py /app/src/
+
+EXPOSE 8000
```
JRegimbal, line 21: You don't actually seem to use port 8000.

```suggestion
EXPOSE 80
```

services/multilang-support/src/translate.py
```diff
@@ -0,0 +1,89 @@
...
+
+    LOGGER.debug("- Request validated! -")
+
+    # Get text to translate
+    segments: list = content["segments"]
+    # source lang is optional, hence the try/except
+    try:
+        source_lang = content["src_lang"]
+    except KeyError:
+        source_lang = "en"
```
JRegimbal, line 62: Suggest adding a debug statement specifying that you're defaulting to English.

services/multilang-support/src/utils.py
```diff
@@ -0,0 +1,221 @@
...
+
+if "utils" in __name__ or __name__ == "__main__":
+    instantiate()
+    LOGGER.debug(f"Using {DEVICE} on {DEVICE_NAME}")
+    LOGGER.info('Service is instantiated and ready.')
+
+
+class Translator:
+    """
+    (Not yet implemented fully)
```
JRegimbal, line 190: What is remaining for this to be implemented?

JRegimbal (changes requested):
Minor changes/questions, but overall quite good. After clarifying on Slack, I do not believe there is a startup/first request timing problem worth delaying integration over.

services/multilang-support/src/translate.py
notkaramel: Added!

services/multilang-support/src/utils.py
notkaramel: My idea is to generalise the service into `Translator` objects for each language pair. Each object will perform tasks like the service right now using their own `MODEL` and `TOKENIZER` instead of a global one. 

Basically, I (or future developers) will have to implement all the functions above but as a class function (with e.g., `self.MODEL`) and calling `Translator('en', 'es').translate(segments)` in translate.py

notkaramel: I think this part can be ignored to finish the sprint/PR. I had a discussion with @rbazin  about ML and using the models, and we figured that there's another approach to implementing the service that is faster and utilize the resources better, since it uses the Marian framework that it was trained on rather than via an API like HuggingFace or Pytorch.

JRegimbal (approved):

Issue #711, opened 2023-06-05
OSM Debug Handler does not handle non-map inputs correctly


JRegimbal:
When sent a photo or other non-map input, the OSM Debug SVG handler does not respond correctly. This is not logged in the handler itself, as it should be, but is caught by the orchestrator.
```
image-orchestrator-1           | FetchError: invalid json response body at http://svg-open-street-map-handler:5000/handler reason: Unexpected end of JSON input
image-orchestrator-1           |     at /usr/src/app/node_modules/node-fetch/lib/index.js:273:32
image-orchestrator-1           |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
image-orchestrator-1           |     at async Promise.all (index 8) {
image-orchestrator-1           |   type: 'invalid-json'
image-orchestrator-1           | }
```

PR #715, opened 2023-06-06
Set debug levels for OSM-debug handler

samueljohnsegun148:
This PR sets logging levels for OSM debug handler and will so handle issue  #711. Prior to now, OSM debug handler would simply check the incoming request. If the request was for a map, it would process, but would however return an invalid JSON response for a non-map case, which is not ideal. 

This is now fixed with the current debug levels implemented. So, in the event of a non-map request, it  won't still process but would return a valid JSON response together with the following debug information:
```23-06-09 22:09 UTC [DEBUG]: Received request
23-06-09 22:09 UTC [DEBUG]: Validating request
23-06-09 22:09 UTC [DEBUG]: Checking whether renderer is supported
23-06-09 22:09 UTC [DEBUG]: Checking for OpenStreetMap map data
23-06-09 22:09 UTC [INFO]: OSM map data not present. Skipping ...
23-06-09 22:09 UTC [DEBUG]: Sending response
```
And if the request is a map type, we would have something similar to this:
```
23-06-09 22:20 UTC [DEBUG]: Received request
23-06-09 22:20 UTC [DEBUG]: Validating request
23-06-09 22:20 UTC [DEBUG]: Checking whether renderer is supported
23-06-09 22:20 UTC [DEBUG]: Checking for OpenStreetMap map data
23-06-09 22:20 UTC [DEBUG]: Map data found! Processing data!
23-06-09 22:20 UTC [DEBUG]: Providing final result!
23-06-09 22:20 UTC [DEBUG]: Sending response
```

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

### DIFF
--- a/handlers/svg-open-street-map/osm-svg.py
+++ b/handlers/svg-open-street-map/osm-svg.py
@@ -36,7 +36,7 @@ LOGGER.setLevel(logging.DEBUG)
 
 @app.route("/handler", methods=["POST"])
 def handle():
-    logging.debug("Received request")
+    LOGGER.debug("Received request")
     # Load necessary schema files
     with open("./schemas/definitions.json") as f:
         definitions_schema = json.load(f)
@@ -58,6 +58,7 @@ def handle():
     # Get and validate the request contents
     contents = request.get_json()
     try:
+        LOGGER.debug("Validating request")
         validator = jsonschema.Draft7Validator(
             request_schema, resolver=resolver
         )
--- a/handlers/svg-open-street-map/osm-svg.py
+++ b/handlers/svg-open-street-map/osm-svg.py
@@ -325,7 +325,7 @@ def handle():
                 LOGGER.debug("Failed to generate a valid response")
                 LOGGER.error(e)
                 return jsonify("Failed to generate a valid response"), 500
-            LOGGER.debug("Finalresponse")
+            LOGGER.debug("Sending final response")
             return response
 
 
--- a/handlers/svg-open-street-map/osm-svg.py
+++ b/handlers/svg-open-street-map/osm-svg.py
@@ -325,7 +325,7 @@ def handle():
                 LOGGER.debug("Failed to generate a valid response")
                 LOGGER.error(e)
                 return jsonify("Failed to generate a valid response"), 500
-            LOGGER.debug("Sending final response ")
+            LOGGER.debug("Sending final response")
             return response
 
 
--- a/handlers/svg-open-street-map/osm-svg.py
+++ b/handlers/svg-open-street-map/osm-svg.py
@@ -136,11 +136,11 @@ def handle():
         if "ca.mcgill.a11y.image.preprocessor.openstreetmap" in preprocessor:
             LOGGER.debug("Openstreetmap (OSM) response found!")
 
-            dt = preprocessor["ca.mcgill.a11y.image.preprocessor.openstreetmap"]
-            if "streets" in dt:
-                streets = dt["streets"]
-                lat = dt["bounds"]["latitude"]
-                lon = dt["bounds"]["longitude"]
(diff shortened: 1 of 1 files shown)


handlers/svg-open-street-map/osm-svg.py
```diff
@@ -75,186 +123,210 @@ def handle():
...
-                                stroke_width=1.5,
-                                stroke='red'))
-                        all_svg.append(
-                            draw.Text(
-                                points_of_interest["cat"],
-                                16,
-                                longitude,
-                                latitude,
-                                fill='black'))
+            svg_layers = []
```
emmanuelwilson, line 153: Some blocks of code like this one does not seem to align with the changes you've described in your PR. Let us know what else has been done in this PR so we can properly review it please! You only mention the change in debug messages (which are quite nice) but it seems like you've done more than just change the debug messages!

samueljohnsegun148: Thanks, but nothing seems to be outside of what was described. The changes that look like some block of codes are not new additions, but appear so with the change of order/organization of the existing blocks brought by the introduction of the debug levels. For an example, in line 132 where the check for map or non-map data is taking place, the existing logic there would simply return an invalid JSON response (i.e. for non-map data), which is one major problem we are addressing with this PR, and it is addressed by the lines 134 -148. This however introduces the supposedly block change observed from line 149. A similar case for the debug level set at line 150 (that confirms if map data is found).

handlers/svg-open-street-map/osm-svg.py
```diff
@@ -48,23 +58,61 @@ def handle():
...
         )
         validator.validate(contents)
     except ValidationError as e:
-        logging.error(e)
-        return jsonify("None"), 204
+        LOGGER.error(e)
+        return jsonify("Invalid request received!"), 400
 
     # Check preprocessor data
+    if "preprocessors" not in contents:
```
JRegimbal, line 71: Why is this separate from the if statement directly below? This is a clear `else` state.

handlers/svg-open-street-map/osm-svg.py
```diff
@@ -75,186 +123,210 @@ def handle():
...
             except jsonschema.exceptions.ValidationError as error:
-                logging.error(error)
+                LOGGER.error(error)
                 return jsonify("Invalid Preprocessor JSON format"), 500
-            logging.debug("Sending response")
+            LOGGER.debug("Sending response")
             return response
 
+        LOGGER.debug("Checking for OpenStreetMap map data ")
         if "ca.mcgill.a11y.image.preprocessor.openstreetmap"\
```
JRegimbal, line 132: Same as above...this is just an else statement for the following if.

JRegimbal (changes requested):
Confusing flow in some places, but otherwise seems good

handlers/svg-open-street-map/osm-svg.py
samueljohnsegun148: Well, I was only being careful not to add things that would be portrayed as new changes.  I've however made the changes now.

JRegimbal (changes requested):

handlers/svg-open-street-map/osm-svg.py
```diff
@@ -325,7 +325,7 @@ def handle():
                 LOGGER.debug("Failed to generate a valid response")
                 LOGGER.error(e)
                 return jsonify("Failed to generate a valid response"), 500
-            LOGGER.debug("Sending response")
+            LOGGER.debug("Finalresponse")
```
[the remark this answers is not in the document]
samueljohnsegun148: Suggested change effected.

JRegimbal (approved):

PR #722, opened 2023-06-15
Improved implementation of the translation `multilang-support` service

notkaramel:
# Improved implementation of the translation service
- This is a continuation of PR #688 and sprint #662
---
## Major Changes:
- Using generalisation to extend the translator capability via `class Translator` instead of Tokenizer/Model global variables in the old implementation
- The format of `src/translate.py` is relatively unchanged.
- Translation process in `src/utils.py` became class's method, i.e., they now have `self` as a parameter.
- More storage will be used as more models are loaded in (~320MB per language pair).
- Resolve remaining TODO in the old implementation: parameter of model generating new tensor (`generate_output_tensor()`)

## Other changes (misc.):
- Added more instruction to `README.md`. Re-organized some sections for clarity
  - Instructions on *How to properly extend the service to support more languages*
- Added documentations from HuggingFace about Tokenizer, Model, and how to use them
- Added flags to `gunicorn` to fix short timeout when downloading multiple models, and to add visual clarity.
- Change `--loglevel` flag to `info` since (tl;dr: to make logs cleaner) the amount of logs printed is the same, with less unwanted background debug log of some process.
- Lots of small visual clarity added
- Removed `@log` decorator in `src/utils.py` because the service no longer need every-step logs. Kept the functionality of `elasped_time` by manually added to `translate()` function.

## Testing
- The new implementation will go through the same testing process as #688.
- Additional tests include:
  - GPU selection handling: different models might be on different GPUs if one is full. They can also fallback to CPU if there is no GPU available.
  - TBD

> Please ensure you've followed the checklist and provide all the required information *before* requesting a review.
> If you do not have everything applicable to your PR, it will not be reviewed!
> If you don't know what something is or if it applies to you, ask!

> Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

### DIFF
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -20,4 +20,4 @@ COPY services/multilang-support/src/*.py /app/src/
 
 EXPOSE 80
 ENV FLASK_APP=src/app.py
-CMD [ "gunicorn", "src.translate:app", "--bind", "0.0.0.0:80", "--log-level=info", "--timeout=120", "--workers=2"]
+CMD [ "gunicorn", "src.translate:app", "--bind", "0.0.0.0:80", "--log-level=info", "--timeout=120"]
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -20,4 +20,4 @@ COPY services/multilang-support/src/*.py /app/src/
 
 EXPOSE 80
 ENV FLASK_APP=src/app.py
-CMD [ "gunicorn", "src.translate:app", "-b", "0.0.0.0:80", "--log-level=debug"]
+CMD [ "gunicorn", "src.translate:app", "--bind", "0.0.0.0:80", "--log-level=info", "--timeout=120", "--workers=2"]
--- a/services/multilang-support/Dockerfile
+++ b/services/multilang-support/Dockerfile
@@ -13,7 +13,14 @@ RUN pip install -r requirements.txt
 
 # Clone the model
 RUN git lfs install
+# English -> French
 RUN git clone https://huggingface.co/Helsinki-NLP/opus-mt-en-fr ./models/opus-mt-en-fr
+
+# English -> German
+RUN git clone https://huggingface.co/Helsinki-NLP/opus-mt-en-de ./models/opus-mt-en-de
+
+# English -> Spanish
+RUN git clone https://huggingface.co/Helsinki-NLP/opus-mt-en-es ./models/opus-mt-en-es
 
 COPY schemas/services/translation.schema.json /app/
 COPY services/multilang-support/src/*.py /app/src/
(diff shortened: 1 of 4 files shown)


services/multilang-support/Dockerfile
```diff
@@ -20,4 +20,4 @@ COPY services/multilang-support/src/*.py /app/src/
 
 EXPOSE 80
 ENV FLASK_APP=src/app.py
-CMD [ "gunicorn", "src.translate:app", "-b", "0.0.0.0:80", "--log-level=debug"]
+CMD [ "gunicorn", "src.translate:app", "--bind", "0.0.0.0:80", "--log-level=info", "--timeout=120", "--workers=2"]
```
notkaramel, line 23: Notes on changes here:
- `--log-level=info`: Python `logging` still allows debug level message to passthrough and printed on the terminal/console of the docker container.
- `--timeout=120`: this helps prevent worker timeout error that occurs when starting, workers take extra time to download models and restarts after one or two models are instantiated, which lead to extremely long boot-up time for the Translator if we support more languages.
- `--workers=2`: a nice feature to have, help processing multiple requests concurrently.

services/multilang-support/src/utils.py
```diff
@@ -9,10 +9,11 @@
 # Constants
 MAX_TIME = 0.5  # [seconds], used for maximum time allowed
 # for translation generation in `generate_output_tensor()`
+SUPPORTED_LANGS = ["fr", "de", "es", "it", "nl", "ru", "zh", "vi"]
```
notkaramel, line 12: Each model takes about ~300MB of storage for its pytorch model weight. Here I have 8 models just as proof of concept that we can easily extend the **multilang** aspect of the service.

notkaramel:
Pre-download models at Docker Build time. This help reduce instantiate time whenever the container is restarted/rebuild. Also take advantage of docker pre-cached files:
image

Also implemented fail safe in case future developer adds language to `SUPPORTED_LANGS` but forgets to pre-download model in Dockerfile, such that it will download directly from HuggingFace.

services/multilang-support/Dockerfile
notkaramel: UPDATE: removed `--workers` flag due to some unexplicable behaviour between them while running the service :)

JRegimbal: Thanks on removing workers, I'd honestly rather handle scaling up at the docker level rather than hard coded into the image itself!

services/multilang-support/README.md
```diff
@@ -32,25 +58,41 @@ from transformers import AutoTokenizer, AutoModelForSeq2SeqLM
...
+    │   ├── target.spm
+    │   └── ... # other files
+    └── # other models (if any)
+        └── ...
 ```
 - `gunicorn` will be called from the `app` repository.
 
+### Extending to other languages
+- To properly extend the service to support other languages, one must follow these steps:
+  - [x] Add the language code (ISO ) to `SUPPORTED_LANGS` list in `src/utils.py`.
```
JRegimbal, line 76:
```suggestion
  - [x] Add the language code (ISO 639-1) to `SUPPORTED_LANGS` list in `src/utils.py`.
```

JRegimbal (approved):
This looks good!

Issue #731, opened 2023-07-03
openstreetmap preprocessor not handling lack of API key appropriately?


jeffbl:
CONFIGURATION ISSUE: @samueljohnsegun148  I'm testing on unicorn using the map on the [tech page of the IMAGE website](https://image.a11y.mcgill.ca/pages/tech.html), and I'm seeing an error in the logs that openstreetmap is not finding an API key. Could you look into fixing any problems with the API key on unicorn?

`image-openstreetmap-1                  | 2023-07-03T11:52:19.877596123Z ERROR:root:Unable to find API Key`

MAIN ISSUE: If it can't find an API key, it should probably just return at that point, but I still see:

`image-openstreetmap-1                  | 2023-07-03T11:52:19.877611813Z ERROR:root:Unable to find Latitude/Longitude `

implying that it is continuing to try and run even if it knows that there is no API key it can use. Suggest this should exit as soon as it knows it cannot find an API key? Is this an appropriate message reflecting what is actually happening? Note that the autour preprocessor is returning results, so I'm assuming that the lat/lng are legitimate from the request, so there may be something else going on here?

While revising, please review debug logging and error checking overall - I cannot test it since it doesn't seem to be working due to the API key issue.

Cybernide:
Have been told that this is awaiting a pull request from @JRegimbal - could you please confirm?

samueljohnsegun148:
@Cybernide Yes, two pull requests (i.e., #732 and the other in the repo for IMAGE-website), to address this, are awaiting approval. I decided not to directly link this issue since they are two separate pull requests and would probably be merged sequentially.

jeffbl:
@samueljohnsegun148 I see #732 has been merged, and I don't see another pending PR. Is this completed? If so, can it be closed?

samueljohnsegun148:
@jeffbl Yes. It can be closed.

Cybernide:
@jeffbl Closing

PR #732, opened 2023-07-05
Update and set additional logging configurations for OSM preprocessor

samueljohnsegun148:
Below is the summary of the changes made:
- Logging is set to debug level
- Changed logging response to a preprocessor response JSON format, when an error is caught
- Added new logging for when an incoming request is being validated
- Added new debug messages 

**Result from testing:**
```
23-07-06 06:13 UTC [DEBUG]: Received request
23-07-06 06:13 UTC [DEBUG]: Validating request
23-07-06 06:13 UTC [DEBUG]: Sending final response
```
There are also error-checking loggings, depending on certain decisions.  Overall, this PR provides better diagnostic features for the OSM preprocessor than the current one. 
Previously, we would only have this:
```
| DEBUG:root:Received request
```
**NB:** This only resolves a section of #731 that concerns debug logging and error checking.
## Required Information

## Coding/Commit Requirements

### DIFF
--- a/preprocessors/openstreetmap/config.py
+++ b/preprocessors/openstreetmap/config.py
@@ -8,3 +8,10 @@ defaultServer = "https://pegasus.cim.mcgill.ca/overpass/api/interpreter?"
 
 secondaryServer1 = "https://maps.mail.ru/osm/tools/overpass/api/interpreter"
 secondaryServer2 = "https://overpass-api.de/api/interpreter"
+
+
+# Set names for map servers
+
+map_server_1 = "Pegasus Server"
+map_server_2 = "VK Maps Overpass API Server"
+map_server_3 = "Main Overpass Server"
--- a/preprocessors/openstreetmap/config.py
+++ b/preprocessors/openstreetmap/config.py
@@ -8,10 +8,3 @@ defaultServer = "https://pegasus.cim.mcgill.ca/overpass/api/interpreter?"
 
 secondaryServer1 = "https://maps.mail.ru/osm/tools/overpass/api/interpreter"
 secondaryServer2 = "https://overpass-api.de/api/interpreter"
-
-
-# Set names for map servers
-
-map_server_1 = "Pegasus Server"
-map_server_2 = "VK Maps Overpass API Server"
-map_server_3 = "Main Overpass Server"
--- a/preprocessors/openstreetmap/main.py
+++ b/preprocessors/openstreetmap/main.py
@@ -87,7 +87,7 @@ def get_map_data():
     # Build OpenStreetMap request
     coords = get_coordinates(content)
     if coords is None:
-        error = 'Unable to find Latitude/Longitude'
+        error = 'Unable to find API key'
         LOGGER.error(error)
         response = {
             "request_uuid": request_uuid,
@@ -194,7 +194,7 @@ def get_map_data():
 
     if validated is not None:
         return validated
-    LOGGER.debug("Sending response")
+    LOGGER.debug("Sending final response")
     return response
 
 
--- a/preprocessors/openstreetmap/main.py
+++ b/preprocessors/openstreetmap/main.py
@@ -76,7 +76,7 @@ def get_map_data():
     # Build OpenStreetMap request
     coords = get_coordinates(content)
     if coords is None:
-        error = 'Unable to find API key'
+        error = 'Unable to find latitude/longitude'
         LOGGER.error(error)
         return jsonify(""), 400
 
--- a/preprocessors/openstreetmap/main.py
+++ b/preprocessors/openstreetmap/main.py
@@ -77,13 +77,15 @@ def get_map_data():
             "name": name,
             "data": {}
         }
-        try:
-            validator = jsonschema.Draft7Validator(
-                response_schema, resolver=resolver)
-            validator.validate(response)
-        except jsonschema.exceptions.ValidationError as error:
-            LOGGER.error(error)
-            return jsonify("Invalid Preprocessor JSON format"), 500
+        validated = validate(
+            schema=schema,
+            data=response,
+            resolver=resolver,
+            json_message='Invalid Preprocessor JSON format',
+            error_code=500)
+
+        if validated is not None:
+            return validated
         LOGGER.debug("Sending response")
         return response
 
@@ -98,13 +100,15 @@ def get_map_data():
             "name": name,
             "data": {}
         }
-        try:
-            validator = jsonschema.Draft7Validator(
-                response_schema, resolver=resolver)
-            validator.validate(response)
-        except jsonschema.exceptions.ValidationError as error:
-            LOGGER.error(error)
-            return jsonify("Invalid Preprocessor JSON format"), 500
+        validated = validate(
+            schema=schema,
+            data=response,
+            resolver=resolver,
+            json_message='Invalid Preprocessor JSON format',
+            error_code=500)
+
+        if validated is not None:
+            return validated
         LOGGER.debug("Sending response")
         return response
 
--- a/preprocessors/openstreetmap/main.py
+++ b/preprocessors/openstreetmap/main.py
@@ -19,6 +19,16 @@ from osm_service import (
 app = Flask(__name__)
 app.config['JSON_SORT_KEYS'] = False
 
+# Configure logging settings
+logging.basicConfig(
+    level=logging.DEBUG,
+    format="%(asctime)s [%(levelname)s]: %(message)s",
+    datefmt="%y-%m-%d %H:%M %Z",
+)
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.setLevel(logging.DEBUG)
+
 
 @app.route('/preprocessor', methods=['POST', ])
 def get_map_data():
--- a/preprocessors/openstreetmap/main.py
+++ b/preprocessors/openstreetmap/main.py
@@ -75,22 +75,9 @@ def get_map_data():
 
     # Build OpenStreetMap request
     coords = get_coordinates(content)
-    if coords is None:
-        error = 'Unable to find latitude/longitude'
-        LOGGER.error(error)
-        return jsonify(""), 400
 
     latitude = coords["latitude"]
     longitude = coords["longitude"]
-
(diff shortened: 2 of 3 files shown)


preprocessors/openstreetmap/main.py
```diff
@@ -54,26 +65,58 @@ def get_map_data():
 
     if validated is not None:
```
emmanuelwilson, line 66: You are checking for validated before it is declared! Surprised this didn't produce an error unless I am misinterpreting this. Do you need this check here?

preprocessors/openstreetmap/main.py
```diff
@@ -54,26 +65,58 @@ def get_map_data():
...
         return validated
-
+    time_stamp = int(get_timestamp())
+    name = "ca.mcgill.a11y.image.preprocessor.openstreetmap"
+    request_uuid = content["request_uuid"]
     # Check if this request is for an openstreetmap
     if 'coordinates' not in content and 'placeID' not in content:
-        logging.info("Not map content. Skipping...")
-        return "", 204
+        LOGGER.info("Not map content. Skipping...")
```
emmanuelwilson, line 73: Great warning message! But you aren't exactly skipping are you? I don't know if the policy changed, but you are still sending a response for JSON for this preprocessor. The response has no data in it so you are skipping in a sense, but as far as I see for other preprocessors, when you skip you don't return a JSON. No point in populating the output with empty data! Please correct me if I am wrong or if there has been a change in how to go about this!

preprocessors/openstreetmap/main.py
```diff
@@ -54,26 +65,58 @@ def get_map_data():
...
+            "data": {}
+        }
+        validated = validate(
+            schema=schema,
+            data=response,
+            resolver=resolver,
+            json_message='Invalid Preprocessor JSON format',
+            error_code=500)
+
+        if validated is not None:
```
emmanuelwilson, line 87: Since this check is within the same if statement as when you declare "validated", won't this always pass? Will ever reach the the send response?

emmanuelwilson:
When sending a PR, please fill in the "Project" and "Labels" portions on the right of the screen (under the assignees and reviewers). The "Project" section is particularly important since that is what makes your PR show up on the project board! Labels is just good for sorting what the issue is about at a glance.

preprocessors/openstreetmap/main.py
samueljohnsegun148: It was declared before being called. You can actually find the function in the ```osm_service.py``` file (line 973) and imported into the ```main.py`` file (line 15).

preprocessors/openstreetmap/main.py
samueljohnsegun148: I see that as well that other preprocessors do not return a JSON, but I did that as a preemptive move borrowing a clue from the OSM debug handler (#711, #715), where the orchestrator had to catch the returned output as an ```invalid json```, with the current format. However, this has now been slightly modified as I completely agree with you that it might never reach the send response.

preprocessors/openstreetmap/main.py
```diff
@@ -54,26 +65,26 @@ def get_map_data():
...
+        LOGGER.info("Not map content. Skipping...")
+        return jsonify(""), 204
 
     # Build OpenStreetMap request
     coords = get_coordinates(content)
     if coords is None:
-        error = 'Unable to find Latitude/Longitude'
-        logging.error(error)
-        return jsonify(error), 400
+        error = 'Unable to find API key'
```
JRegimbal, line 79: I am not a fan of this error message. This does not distinguish between the API key not being loaded and an error in getting a response from Google. For example, if we request an invalid place ID, this error will still occur and mislead the debugger.

preprocessors/openstreetmap/main.py
```diff
@@ -54,26 +65,26 @@ def get_map_data():
...
+        LOGGER.error(error)
+        return jsonify(""), 400
```
JRegimbal, line 81: If the API key is not loaded, this is not a client (400) error.

preprocessors/openstreetmap/osm_service.py
```diff
@@ -82,16 +93,16 @@ def get_streets(bbox_coord):
     except Exception:
         try:
             error = 'Primary server not responding, so connecting server 1'
```
JRegimbal, line 95: Suggest logging the actual server names rather than just "primary", "server 1", "server 2". The `config.py` may change (or be updated in the future so servers can be overridden) and these messages will not be nearly as informative.

preprocessors/openstreetmap/osm_service.py
```diff
@@ -180,7 +191,7 @@ def process_streets_data(OSM_data, bbox_coordinates):
                     processed_OSM_data.append(way_object)
     except AttributeError:
         error = 'Overpass Attibute error. Retry again'
-        logging.error(error)
+        LOGGER.debug(error)
```
JRegimbal, line 194: Suggest including a backtrace of this error (or similar details).

preprocessors/openstreetmap/osm_service.py
```diff
@@ -661,16 +672,16 @@ def get_amenities(bbox_coord):
     except Exception:
         try:
             error = 'Primary server not responding, so connecting server 1'
```
JRegimbal, line 674: Same comment on including actual server names.

preprocessors/openstreetmap/osm_service.py
```diff
@@ -1008,7 +1018,7 @@ def get_coordinates(content):
     place_response = requests.get(request).json()
 
     if not check_google_response(place_response):
-        return None
+        return jsonify("Processing Aborted!"), 400
```
JRegimbal, line 1021: This is not a client error.

JRegimbal: Please improve the debugging message here as well. "Processing Aborted" does not indicate the cause.

JRegimbal (changes requested):

preprocessors/openstreetmap/main.py
emmanuelwilson: @samueljohnsegun148 for a list of errors and what they mean, I recommend looking at: 
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
You can pick and choose the error # you deem is the correct one!

samueljohnsegun148: @emmanuelwilson, I think I've fixed this already. If you look at the logging message, it is actually about a missing coordinate( latitude/longitude), which should be obviously a client error, but I stand to be corrected.

samueljohnsegun148:
@emmanuelwilson @JRegimbal Please note that all comments have now been addressed.

preprocessors/openstreetmap/main.py
JRegimbal: You check if coordinates were not specified above. Please specify if/when `get_coordinates` will return `None` here.

preprocessors/openstreetmap/osm_service.py
```diff
@@ -660,17 +675,19 @@ def get_amenities(bbox_coord):
         amenities = server_config2(defaultServer, bbox_coord)
     except Exception:
         try:
-            error = 'Primary server not responding, so connecting server 1'
-            logging.error(error)
+            error = 'Pegasus Server not responding, so connecting VK Maps'\
```
JRegimbal, line 678: Please do not hard code the names of servers. This will just lead to a headache later on.

JRegimbal (changes requested):

preprocessors/openstreetmap/osm_service.py
samueljohnsegun148: All the necessary changes were effected.

preprocessors/openstreetmap/config.py
```diff
@@ -8,3 +8,10 @@
 
 secondaryServer1 = "https://maps.mail.ru/osm/tools/overpass/api/interpreter"
 secondaryServer2 = "https://overpass-api.de/api/interpreter"
+
+
+# Set names for map servers
+
+map_server_1 = "Pegasus Server"
```
JRegimbal, line 15: I really don't understand the benefit of using these names when we would need to look here to find out the actual URL anyway.

preprocessors/openstreetmap/main.py
JRegimbal: @samueljohnsegun148 when will this condition actually be met?

preprocessors/openstreetmap/main.py
```diff
@@ -54,26 +65,36 @@ def get_map_data():
...
-        logging.error(error)
-        return jsonify(error), 400
+        error = 'Unable to find latitude/longitude'
+        LOGGER.error(error)
+        return jsonify(""), 400
 
     latitude = coords["latitude"]
     longitude = coords["longitude"]
+
+    if latitude is None:
```
JRegimbal, line 86: If `coordinates` is present, and you validate the request, this will never actually come up.

JRegimbal (changes requested):

preprocessors/openstreetmap/config.py
samueljohnsegun148: @JRegimbal, do you have any suggestions in mind on this?

preprocessors/openstreetmap/main.py
JRegimbal: So then what does this whole branch of checking if `coords` is `None` actually do?

preprocessors/openstreetmap/config.py
JRegimbal: I'd just directly use URLs in log messages.

preprocessors/openstreetmap/main.py
samueljohnsegun148: As a matter of fact, this as well as the logic in ```line 78```  of the ```main.py``` is all redundant and shouldn't be there anymore. Thanks for pointing it out.

preprocessors/openstreetmap/config.py
samueljohnsegun148: Okay, that has been done now.

JRegimbal (approved):

Issue #723, opened 2023-06-16
French TTS (`espnet-tts-fr`) struggles with numerical numbers


notkaramel:
As title suggests, the French TTS is having a hard time generating audio for large number input, in numeric form. Here in this example, the text output from the translation is (from [Etherscan](https://etherscan.io/chart/ethersupplygrowth)): 

- "Graphique de croissance de l'offre d'Ether. x Axis,Time,du 30 juillet 2015 au 15 juin 2023. y Axis,Ether Supply,du 72049306.59375 au 120532870.54151082"

Here is the audio generated: [French TTS Rendering (via Google Drive)](https://drive.google.com/file/d/18nrvcSZLEdWy5ThsalGDsdlzF0JgH6BP/view?usp=sharing).

Notice that the service struggles to say "2015", "2023", and most of the remaining large numbers.

notkaramel:
French TTS also struggles with describing data. Example from a pie chart from https://www.inspq.qc.ca/covid-19/donnees/age-sexe

![image](https://github.com/Shared-Reality-Lab/IMAGE-server/assets/24505220/e729dce8-fb1f-4cec-b268-aae589790110)

Audio: [Pie Chart Rendering (via Google Drive)](https://drive.google.com/file/d/12rpbvbkSSlyMhD_8hPZNM56a7gZqbPDU/view?usp=sharing)

Note that the French TTS struggles to say concatenated numbers or with decimal point/comma, like (in **bold**):
- "**40-49** ans, **15,76** pour cent"
- "**50-59** ans, **12,70** pour cent"
- ...

notkaramel:
Weird case: French TTS can not say "2" or "3" in French.

Graphic:
![Image from IMAGE](https://image.a11y.mcgill.ca/pages/imgs/pexels-mikhail-nilov-6981101.jpg)

Text (in French): "Cette photo intérieure contient les objets ou personnes suivants: 3 livres, 3 téléviseurs, 2 personnes, 2 chaises, un téléphone portable, une souris, et un clavier."

Audio: [Photo rendering (via Google Drive)](https://drive.google.com/file/d/1X225D5rOorS-zRVYr6ITGufCyan63iKI/view?usp=sharing)

Notice how French TTS said "3" as "et tres" (I think), and "2" as "two". "2 personnes" should have been "deux personnes" but it was "two personnes".

jeffbl:
[...]

PR #739, opened 2023-07-19
Helping French TTS to pronounce numbers 

notkaramel:
## Changes Made 
- Issue: #723 
- Added `num2words` library to help convert numbers in numeric form to text (e.g. "42" to "quarante deux")
- Added debugging logs indicating cases of numeric conversion: ("segment" as a part of the input string that's separated by a space)
  - Case 0: segment is numeric or float number (e.g., "2012", "23.12", "2021."). This includes case where a number is at the end of a sentence with "." as a part of the segment.
  - Case 1: segment has ",". This is due to `multilang-support` tranlated to Metropolitan French where they have a different notation system.
  - Case 2: segment is a range, containing a hyphen '-'. This is often seen in charts with age or year range. It also requires extra syntax to be added. E.g., "34-59" would be "**de** trente quatre **à** cinquante neuf". 
  - > **NOTE:** This might conflict with words in French that has hyphen, like "gratte-ciel", "arc-en-ciel", or names like "Lionel-Groulx".

## Testings
- Tested with large numbers with demimal point, year: https://etherscan.io/chart/ethersupplygrowth
- Tested with range: https://www.inspq.qc.ca/covid-19/donnees/age-sexe
- Tested with regular numbers: any graphics with countable quantity of object(s). 

---

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

### DIFF
--- a/services/espnet-tts-fr/src/app.py
+++ b/services/espnet-tts-fr/src/app.py
@@ -129,7 +129,6 @@ def segment_tts():
                     segment_new.append(s)
 
             segment_new = " ".join(str(s) for s in segment_new)
-            logger.debug(f'New Segment: {segment_new}')
             wavs.append(tts(segment_new))
         for wav in wavs:
             if totalWav is not None:
--- a/services/espnet-tts-fr/src/app.py
+++ b/services/espnet-tts-fr/src/app.py
@@ -50,22 +50,19 @@ def isfloat(num):
 
 
 def processSegment(s):
-    if s.isdigit() or isfloat(s):
+    if s.isnumeric() or isfloat(s):
+        logger.debug(f"Case: '{s}' is numeric or float.")
         return num2words(s, lang='fr')
-    if "." in s:
-        temps = s.replace(".", "")
-        if temps.isnumeric():
-            temps = int(temps)
-            s = num2words(int(temps), lang='fr')
-    if "," in s:
+    elif "," in s:
+        logger.debug(f"Case: '{s}' has , as separator")
         tempns = s.replace(",", ".")
         if isfloat(tempns):
-            s = num2words(tempns, lang='fr')
-    if "-" in s:
+            return num2words(float(tempns), lang='fr')
+    elif "-" in s:
+        logger.debug("case 3: -")
         num_in_ns = s.split("-")
-        s = " ".join(["de", num2words(num_in_ns[0], lang='fr'),
-                      "à", num2words(num_in_ns[1], lang='fr')])
-    return s
+        return " ".join(["de", num2words(num_in_ns[0], lang='fr'),
+                         "à", num2words(num_in_ns[1], lang='fr')])
 
 
 @app.route("/service/tts/simple", methods=["POST"])
@@ -116,8 +113,14 @@ def segment_tts():
             segment_new = []
             logger.debug(segment.split())
             for s in segment.split():
-                logger.debug(f'Performing on: {s}')
-                segment_new.append(processSegment(s))
+                try:
+                    logger.debug(f'Performing on: "{s}"')
+                    segment_new.append(processSegment(s))
+                except Exception as e:
+                    logger.error("ERROR processing")
+                    logger.error(e)
+                    segment_new.append(s)
+
             segment_new = " ".join(segment_new)
             logger.debug(f'New Segment: {segment_new}')
             wavs.append(tts(segment_new))
--- a/services/espnet-tts-fr/src/app.py
+++ b/services/espnet-tts-fr/src/app.py
@@ -27,10 +27,15 @@ from jsonschema import validate
 from torch.cuda import empty_cache
 from werkzeug.wsgi import FileWrapper
 from num2words import num2words
+import re  # for regular expression processing
 
 logging.basicConfig(format="%(asctime)s %(message)s")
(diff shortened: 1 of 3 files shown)


services/espnet-tts-fr/src/app.py
```diff
@@ -40,6 +41,30 @@
 app = Flask(__name__)
 
 
+def isfloat(num):
+    try:
+        float(num)
+        return True
+    except ValueError:
+        return False
```
notkaramel, line 49: I added this because there's (supposedly) no built-in function in Python to check if a number is a float.

services/espnet-tts-fr/src/app.py
```diff
@@ -82,7 +107,23 @@ def segment_tts():
...
         durations = []
-        wavs = [tts(segment) for segment in data["segments"]]
+        wavs = []
+        for segment in data["segments"]:
+            # detect numerical in segments
+            segment_new = []
+            logger.debug(segment.split())
+            for s in segment.split():
+                try:
+                    logger.debug(f'Performing on: "{s}"')
```
JRegimbal, line 117: This is going to get _very_ chatty very quickly...

JRegimbal (approved):
Looking great!

services/espnet-tts-fr/src/app.py
JRegimbal: Not merging right now, but do you anticipate this being useful going forward? I'm a little worried about the slim possibility of this leaking some personal info.

notkaramel: I find this helpful to debug what component the TTS couldn't convert to numbers sometimes, but it can definitely be case specific (log in the `except` statement) rather than logging out everything. 

You might be right... In that case, I would have to change the logs of `multilang-support` as well because it has been logging the whole segment out everytime, both in English and French 🫤

JRegimbal: Ah, yeah please do that too then!

notkaramel: Done that!

notkaramel:
@JRegimbal I just reformat the `processSegment` so that it would let normal text through + added edge case handling for hyphen cases: it can now handle number range and also words/names like "gratte-ciel" (skyscrapper), "Lionel-Groulx", etc., without trying to convert them to number :)

services/espnet-tts-fr/src/app.py
```diff
@@ -51,18 +51,26 @@ def isfloat(num):
...
     elif "," in s:
-        logger.debug(f"Case: '{s}' has , as separator")
+        logger.debug("Case: has a , as separator")
         tempns = s.replace(",", ".")
         if isfloat(tempns):
-            return num2words(float(tempns), lang='fr')
+            s = num2words(float(tempns), lang='fr')
     elif "-" in s:
-        logger.debug("case 3: -")
+        logger.debug("Case: has a - as separator")
```
JRegimbal, line 62: Hmm, this would likely fail for cases where two separators are used. E.g., "39,5-40,0". We could probably do something with regexps with the - separator as well. Get the fields we need excluding text with `([\d,.]+)-([\d,.]+)`, replace with `de $1 à $2` and then apply another find and replace on $1 and $2 to replace occurrences of `,` with `.`. Does this make sense?

services/espnet-tts-fr/src/app.py
```diff
@@ -51,18 +51,26 @@ def isfloat(num):
 
 def processSegment(s):
     if s.isnumeric() or isfloat(s):
-        logger.debug(f"Case: '{s}' is numeric or float.")
-        return num2words(s, lang='fr')
+        logger.debug("Case: numeric or float.")
+        s = num2words(s, lang='fr')
     elif "," in s:
-        logger.debug(f"Case: '{s}' has , as separator")
+        logger.debug("Case: has a , as separator")
```
JRegimbal, line 57: Hmm, does this still get run for text uses of `,`?

JRegimbal: Could also catch with a regexp find replace (`(\d+),(\d+)` to `$1.$2`).

JRegimbal (changes requested):

services/espnet-tts-fr/src/app.py
notkaramel: It catches text with `,`, but it won't perform any changes since on the next line, it couldn't cast the text to float. 

I will look into regex and string processing more, thanks for the suggestion! I genuinely forgot about regex and went straight for the simpliest implementation.

services/espnet-tts-fr/src/app.py
notkaramel: Yes, it does. I will try them out.

notkaramel:
- @JRegimbal I added regex detection & processing and it's now working great on the examples from #723. Please take a look at the changes!
- One thing I'm concerning is that the TTS doesn't have pauses, user might feel like it's too fast. Should I create a new work item for that?

JRegimbal (approved):
Looks good!

JRegimbal:
@notkaramel you can make a new issue, but I'd put it in backlog for right now and bring it up tomorrow if you're still looking for something to do. There may be a higher priority issue.

PR #781, opened 2023-10-02
Add settable PREPROCESSOR_TIMEOUT env var

### LINKED ISSUE #775: Set orchestrator priority group timeout with an environment variable
I could've sworn I already implemented this, but it seems that there is no actual way to configure the time the orchestrator will wait for a preprocessor to return. Obviously, this should be fixed.

https://github.com/Shared-Reality-Lab/IMAGE-server/blob/9555f3f1d14d8d3bb25f444a3e3acce001561841/orchestrator/src/server.ts#L38C36-L38C36

JRegimbal:
Resolve #775 by adding the `PREPROCESSOR_TIMEOUT` variable which takes the group/individual preprocessor timeout in milliseconds. Defaults to 15 s if not specified. Tested on unicorn by running with it unset (resulting in normal behaviour) and set to 10 (resulting in all preprocessors failing to return and connections being aborted).

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/orchestrator/src/server.ts
+++ b/orchestrator/src/server.ts
@@ -35,7 +35,7 @@ const ajv = new Ajv2020({
     "schemas": [definitionsJSON, querySchemaJSON, responseSchemaJSON, handlerResponseSchemaJSON, preprocessorResponseSchemaJSON]
 });
 
-const PREPROCESSOR_TIME_MS = 15000;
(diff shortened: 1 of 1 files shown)


jeffbl:
Looks fine to me from inspection. I'd argue it could be useful to have timeouts per preprocessor set in the docker-compose.yml (and a global per-group timeout?), but this is much better than having it non-configurable, and is a simple way to change it across the board. Looks good to merge.

Issue #798, opened 2024-01-26
Dynamically generate docker-compose.yml for different servers, based on canonical copy


jeffbl:
PROBLEM: Right now, we have completely separate docker-compose.yml files for [unicorn](https://github.com/Shared-Reality-Lab/IMAGE-website/blob/main/docker-compose/unicorn-docker-compose.yml) and [pegasus](https://github.com/Shared-Reality-Lab/IMAGE-website/blob/main/docker-compose/pegasus-docker-compose.yml), that must manually be kept in sync. Then, separately, we have a "canonical" [example docker-compose.yml in the server repo](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/docker-compose.yml) that also needs to be manually kept in sync, as the one that we expose publicly for people wanting to run their own IMAGE server. This is error-prone due to all of the manual changes necessary, and makes it very difficult for people to know how to integrate their preprocessors/handlers into test and then production. The bulk of the files are all the same, with tweaks for what version of each to run, and some specifics around splitting preprocessors between GPUs on different machines.

GOAL: The canonical docker-compose.yml should be the only place most changes/additions need to be made.

METHOD: Through a combination of [environment variables](https://docs.docker.com/compose/environment-variables/set-environment-variables/), [docker-compose variables](https://stackoverflow.com/questions/48053286/declaring-variables-in-docker-compose), [profiles](https://docs.docker.com/compose/profiles/), and overrides, it seems like this could be done quite cleanly.

BENEFITS:
- clear documentation on where/how to add preprocessors/handlers
- specific configurations for new servers become concise and less error-prone
- If we can parameterize things like the test/production version of the containers being used, then we can realistically and quickly do something like put unicorn into a production mode temporarily if we have problems with pegasus
- The canonical version of the docker-compose becomes a living file that others can realistically use to bring up their own server, and it is more likely to actually work since we'd be using it ourselves. Note that we may need to add documentation/support files to explain how to do this, and document what variables and such need to be set.

Please have @jeffbl review when complete

jeffbl:
@jaydeepsingh25 can you please take a pass through the documentation on the IMAGE-server wiki and update anything related to this, before closing?

PR #808, opened 2024-02-13
fix: single docker compose file

jaydeepsingh25:
The PR includes changes required for https://github.com/Shared-Reality-Lab/IMAGE-server/issues/798. Following is the summary of changes:

- Servers Pegasus and Unicorn will have their own .env files, which will override the default variables.
- There are three different profiles - unicorn, pegasus and default. 
- Important : It is necessary to have profile: `default `for every new service we add . Missing profile field will by run the service for all profiles.
- Command to run containers on Unicorn: `docker compose  --env-file ./unicorn.env up -d`
- Command to run containers on Pegasus: `docker compose  --env-file ./pegasus.env up -d`

@jeffbl , let me know your thoughts on this implementation. I think this should work well. Tested `default `and `unicorn `profiles

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/config/Readme.md
+++ b/config/Readme.md
@@ -2,11 +2,13 @@
 
 This Folder (Config) should contain the .env files required for IMAGE server preprocessors to run precisely.
 
+``` Note: Some preprocessors may require other config as well. Refer to 'Environment setup' section inside Readme of each preprocessor for updated documentation.```
+
 ## Environment Files
 
 ### maps.env
 This file contains the API key used to call Google Places API. [Here](https://developers.google.com/maps/documentation/places/web-service/get-api-key) is the documentation for how to obtain a valid API key. This env file is required by the following preprocessors:
-* [autor preprocessor](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/preprocessors/autour)
+* [autour preprocessor](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/preprocessors/autour)
 * [openstreetmap preprocessor](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/preprocessors/openstreetmap)
 
 Following is the sample format of maps.env file:
@@ -15,7 +17,7 @@ GOOGLE_PLACES_KEY = [INSERT KEY STRING]
 ```
 
 ### apis-and-selection.env
-This env file is required by the following preprocessors:
+Entries in this env file are required by the following preprocessors:
 * [ocr-clouds-preprocessor](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/preprocessors/ocr)
 
 Following is the sample format of apis-and-selection.env file:
@@ -24,7 +26,7 @@ Following is the sample format of apis-and-selection.env file:
 AZURE_API_KEY = [INSERT KEY STRING]
 FREEOCR_API_KEY = [INSERT KEY STRING]
 GOOGLE_APPLICATION_CREDENTIALS = [INSERT KEY FILE PATH AS STRING]
-CLOUD_SERVICE = [INSERT OPTION STRING (see options above)]
+CLOUD_SERVICE = [INSERT OPTION STRING (see options below)]
 ```
 * `CLOUD_SERVICE` determines the desired cloud service to be used. Its possible values are:
     * `AZURE_OCR` (for [Microsoft Azure OCR API](https://westus.dev.cognitive.microsoft.com/docs/services/computer-vision-v3-2/operations/56f91f2e778daf14a499f20d))
@@ -34,6 +36,8 @@ CLOUD_SERVICE = [INSERT OPTION STRING (see options above)]
 * `AZURE_API_KEY` is found in your [Azure portal](https://portal.azure.com)
 * `FREEOCR_API_KEY` can be obtained at [OCR API portal](https://ocr.space/ocrapi)
 * `GOOGLE_APPLICATION_CREDENTIALS` contains path to credentials file. Refer [documentation](https://cloud.google.com/docs/authentication/application-default-credentials#GAC) for details.
+
+
 ### azure-api.env
 This env file is required by the following preprocessors:
 * [graphic tagger](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/preprocessors/graphic-tagger)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -384,7 +384,7 @@ volumes:
 networks:
 # start - reverse proxy setup
 # traefik is a reverse proxy used, any other reverse proxy (nginx, HAProxy) can be used as well  
-# following section is not required if the server is hosted on a normal webserver.
+# following section is not required if the server is hosted in another way
   traefik:
     external: true
     name: traefik
--- a/prod-docker-compose.yml
+++ b/prod-docker-compose.yml
@@ -23,7 +23,7 @@ services:
     environment:
       - SERVER_NAME=image.a11y.mcgill.ca
     volumes:
-      # Following volume mounts are used only by IMAGE website, not required for IMAGE server 
+      # Following volume mounts are used only by the IMAGE project's website, not required for IMAGE server 
       - ./www/image.conf:/etc/nginx/templates/image.conf.template:ro
       - ./www/html:/var/www/html:ro
       - website-logs:/var/log/nginx
--- a/test-docker-compose.yml
+++ b/test-docker-compose.yml
@@ -15,7 +15,7 @@ services:
     environment:
       - SERVER_NAME=unicorn.cim.mcgill.ca
     volumes:
-      # Following volume mounts are used only by IMAGE website, not required for IMAGE server 
+      # Following volume mounts are used only by the IMAGE project's website, not required for IMAGE server 
       - ./auditory-haptic-graphics-website/html:/var/www/html:ro
       - ./auditory-haptic-graphics-website/image.conf:/etc/nginx/templates/image.conf.template:ro
       - ./auditory-haptic-graphics-website/passwords:/etc/nginx/passwords
--- a/preprocessors/autour/README.md
+++ b/preprocessors/autour/README.md
@@ -4,5 +4,14 @@ The autour preprocessor is used to analyze an embedded google map and return loc
 
 This uses the same back-end server as the Autour app (https://autour.mcgill.ca).
 
+## Environment setup
+The environment file (maps.env) should contain the API key used to call Google Places API. [Here](https://developers.google.com/maps/documentation/places/web-service/get-api-key) is the documentation for how to obtain a valid API key.: 
+
+Following is the sample format of maps.env file:
+```
+GOOGLE_PLACES_KEY = [INSERT KEY STRING]
+```
+
 ## Example Website:
 https://developers.google.com/maps/documentation/embed/embedding-map
+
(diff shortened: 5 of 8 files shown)


.gitignore
```diff
@@ -1,7 +1,8 @@
 .DS_Store
 __pycache__/
 *.swp
-.env
+pegasus.env
```
jeffbl, line 4: Since these don't contain any secrets, would it be ok to check them in? (With a comment inside reminding people not to put any secrets inside!) If we don't add these, should probably at least have an example one for the default case. The concern is that we want it to be as clear as possible to someone bringing up their own server what they need to add, variables they need to set, etc.

docker-compose.yml
```diff
@@ -1,306 +1,441 @@
...
-            ca.mcgill.a11y.image.handler: enable
-    map-tactile-svg-handler:
-        image: ghcr.io/shared-reality-lab/image-handler-map-tactile-svg:unstable
-        labels:
-            ca.mcgill.a11y.image.handler: enable
-    svg-action-recognition-handler:
-        image: ghcr.io/shared-reality-lab/image-handler-svg-action-recognition:unstable
-        labels:
-            ca.mcgill.a11y.image.handler: enable
+  image-unicorn-cim-mcgill-ca:
```
jeffbl, line 4: Since this and the pegasus one (and goaccess) are only for running the website, and have nothing to do with running a working IMAGE server, would it be cleaner to remove these from the core docker-compose.yml? This is likely to only cause confusion with external people trying to understand how to run a server. Can be broken out into a separate issue if you want to deal with it separately.

docker-compose.yml
```diff
@@ -1,306 +1,441 @@
...
+        resources:
+          limits:
+            memory: 2g
+
+  graphic-tagger:
+    profiles: [pegasus, unicorn, default]
+    image: ghcr.io/shared-reality-lab/image-preprocessor-graphic-tagger:${REGISTRY_TAG}
+    restart: unless-stopped
+    env_file:
+      - ./config/azure-api.env
```
jeffbl, line 156: Since these other configs are not checked in, likely need some documentation or an example file, so people know what it needs to contain?

jeffbl (changes requested):
This is super exciting, and should be much clearer!

- Anyone outside of our team will not know what "pegasus" and "unicorn" profiles mean. Suggest "production" and "test" to make this generic.
- Would it be better to just have a soft link to the appropriate .env file, rather than manually specify it on the command line when bringing up the containers? This will cause confusion if people forget to do this each time, e.g., when testing on unicorn?
- Few other comments inline, some of which are not necessarily in scope of this specific issue, but address other potential confusions if someone is trying to run their own server. Feel free to break any out into separate issues if you think that is a better way to handle.

ping to @JRegimbal in case you want to have a look before revisions, or have opinions on the above modifications.

docker-compose.yml
```diff
@@ -1,306 +1,441 @@
...
+    restart: unless-stopped
+    labels:
+      ca.mcgill.a11y.image.preprocessor: 3
+      ca.mcgill.a11y.image.port: 5000
+    deploy:
+          resources:
+            reservations:
+              devices:
+              - driver: nvidia
+                device_ids: ['1']
```
jeffbl, line 173: Won't this fail on pegasus, since it only has a single GPU? Need a way to override this based on generic docker-compose for differences in things like GPU hardware (unicorn has two GPUs). Suggest default being the pegasus single card scenario, and an override activated for unicorn, where we need to split specific preprocessors onto second GPU?

docker-compose.yml
jeffbl: Looking again now, I am wrong on this, since these traefik sections also do the routing of all of the preprocessor requests, etc, correct? So it might still make sense to break these out, but they do rely on orchestrator, etc. so they are intertwined. I'm not sure which way to go on this. Is it worth separating them out as separate docker-compose files, or is this more trouble than it is worth since they are in fact connected? How would we want this to be segmented if we were an external team bringing up our own server, for which we might not want a website at all? Would it be clear what we should do, or is there in fact a better way to organize this?

docker-compose.yml
jeffbl: If we knew that docker-compose would assign them in a reasonable way just always using "all", we wouldn't have to deal with this at all. But I don't know how it does the assignments in that case. So, I assume we want the "default" to be "all" (which should work fine on Pegasus since it only has one GPU) and would be a reasonable default for most external configs. Your approach here makes sense, then, even if it seems messy. The only other way that comes to mind would be to make a docker-compose.override.yml file that only changes the GPU assignments. But then, our standard operating procedure for having people create their own override when testing on unicorn becomes more difficult. Multiple overrides are supported, but would need to be manually specified, or we'd have to create nice helper scripts that do the right thing. Not sure if that is worth it. @JRegimbal suggested another .env variable, but this seems weird to me since it is tightly linked to specific preprocessors. Would that really be cleaner? I'm open to debate on this, or to another alternative... I'm fine with the current solution you've implemented for now, and we can revisit later. In speaking with Juliette, one consideration is what will scale better if we move to something like docker swarm, but I don't know since I've never used swarm...

jeffbl:
Assign to @JRegimbal to look this over before we do another round of implementation. Goal is to get this checked in since I think it is much better than what we had before, even if we know there are some rough edges.

config/azure-api.env
```diff
@@ -0,0 +1 @@
+AZURE_API_KEY = ""
```
JRegimbal, line 1: There need to be comments in these env files adding instructions for how to fill in these values for non-IMAGE people.

docker-compose.yml
JRegimbal: I don't think that this is how gitignore works...I think that once a file is tracked, the ignore is not applied to it. Hope I'm wrong here though, since this would be an ideal solution.

docker-compose.yml
JRegimbal: I don't love this, but I'm fine with the two versions in the meantime if there's a comment clearly explaining what's happening.

production.env
```diff
@@ -0,0 +1,5 @@
+# Do not add any secrets in this file
+COMPOSE_PROFILES = production
+COMPOSE_FILE = docker-compose.yml:prod-docker-compose.yml
+REGISTRY_TAG = latest
+DOCKER_GID = 999
```
JRegimbal, line 5: I don't believe this is correct?

test.env
```diff
@@ -0,0 +1,5 @@
+# Do not add any secrets in this file
+COMPOSE_PROFILES = test
+COMPOSE_FILE = docker-compose.yml:test-docker-compose.yml
+REGISTRY_TAG = unstable
+DOCKER_GID = 999
```
JRegimbal, line 5: I also don't believe this is correct.

config/Readme.md
```diff
@@ -0,0 +1,40 @@
...
+This Folder (Config) should contain the .env files required for IMAGE server preprocessors to run precisely.
+
+## Environment Files
+
+### maps.env
+This env file is required by the following preprocessors:
+* [autor preprocessor](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/preprocessors/autour)
+* [openstreetmap preprocessor](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/preprocessors/openstreetmap)
+
+Following is the sample format of maps.env file:
```
JRegimbal, line 12: Please explain what the key is (e.g., how someone would obtain a valid API key). This can just be a link.

config/Readme.md
```diff
@@ -0,0 +1,40 @@
...
+```
+GOOGLE_PLACES_KEY = [INSERT KEY STRING]
+```
+
+### apis-and-selection.env
+This env file is required by the following preprocessors:
+* [ocr-clouds-preprocessor](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/preprocessors/ocr)
+
+The desired cloud service is determined by an environment variable called CLOUD_SERVICE, it is a string and the possible values are:
```
JRegimbal, line 21: Again: explain where these API keys need to come from, and especially the difference between `AZURE_OCR` and `AZURE_READ`.

docker-compose.yml
```diff
@@ -1,306 +1,389 @@
...
+# end - unicorn exclusive services        
 volumes:
-    sc-store:
-    user-logs:
+  sc-store:
+  user-logs:
+  website-logs:
+# https://docs.docker.com/compose/networking/
+networks:
+  traefik:
```
JRegimbal, line 385: I'm not sure about including traefik here, as this is sort of a specific thing we do...maybe include a note on _why_ it's here and how it can be removed/replaced? Or move to the prod/test yml files? Thoughts, @jeffbl?

prod-docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+      - "traefik.http.routers.image.tls.certresolver=myresolver"
+      - traefik.docker.network=traefik
+    environment:
+      - SERVER_NAME=image.a11y.mcgill.ca
+    volumes:
+      - ./www/html:/var/www/html:ro
+      - ./www/image.conf:/etc/nginx/templates/image.conf.template:ro
+      - website-logs:/var/log/nginx
+      - ./www/passwords:/etc/nginx/passwords
+      - ./resources/current:/var/www/resources:ro
```
JRegimbal, line 30: Maybe include a note that these local mounts refer to files used to manage our website, and are unrelated to the "IMAGE server" stuff here. I can see someone getting confused by this.

test-docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+      - "traefik.http.routers.image-unicorn.middlewares=image-unicorn@docker"
+      - traefik.docker.network=traefik
+    environment:
+      - SERVER_NAME=unicorn.cim.mcgill.ca
+    volumes:
+      - ./auditory-haptic-graphics-website/html:/var/www/html:ro
+      - ./auditory-haptic-graphics-website/image.conf:/etc/nginx/templates/image.conf.template:ro
+      - ./auditory-haptic-graphics-website/passwords:/etc/nginx/passwords
+      - ./resources/current:/var/www/resources:ro
+      - website-logs:/var/log/nginx
```
JRegimbal, line 22: Same potential for confusion as the corresponding production block.

JRegimbal (changes requested):

docker-compose.yml
jeffbl: @JRegimbal I agree this is awkward, since it isn't really the IMAGE server stack itself, yet I do think we should have a working example, even if it does only work with one specific reverse proxy (traefik). Perhaps we should add comments to this indicating that traefik is only one reverse proxy, and this can be removed if the server is hosted on a "normal" webserver directly? Again, my preference is to have a fully working example, with (as you suggest) comments to explain potential alternatives. Doing a separate "vanilla" version is unlikely to be kept up-to-date, so may be even more confusing in the long run.

prod-docker-compose.yml
jeffbl: I originally thought we should potentially pull everything to do with the website out of the core docker-compose and put it in another directory with its own docker-compose, but the web front end is kind of intermixed with the endpoints for server requests. Is there a clean way to separate these, that would help with understanding? If unclear, I think documentation in place is the best approach, and people can just remove the tagged lines for the website itself if they only want to serve requests.

jeffbl:
Looks like we're getting close, and I would like to get this merged before next phase of project begins. @jaydeepsingh25 couple more comments above, then unliess @JRegimbal has issues, I think we should get this merged and deployed so that we can work out kinks asap.

JRegimbal (approved):
The changes since the last review look good to me!

jeffbl (approved):

jeffbl:
Still worried we won't remember to keep config/readme.md in sync as we change preprocessors, but let's get this in and take a fresh look in a future sprint, as I expect other tweaks will arise as we change the preprocessor stack during upcoming grant. Thanks!

PR #817, opened 2024-04-28
Memcached impl

jaydeepsingh25:
This PR includes Memcahed Cache implementation at the image server - [issue](https://github.com/orgs/Shared-Reality-Lab/projects/2/views/1?filterQuery=assignee%3Ajaydeepsingh25&pane=issue&itemId=52531261 ). Following are the notes on implementation:

- Cache Memory is 4GB default value as specified in the docker compose. Need to discuss if it should be decreased.
- CacheTimeout is configured at the preprocessor level, with the label `ca.mcgill.a11y.image.cacheTimeout` . Timeout Value of 0 indicates that Cache is disabled for a preprocessor.
- Missing CacheTimeout Label on the preprocessor service will default to timeout value of 0.
- Cache Hashkey is generated with the help of imageBlob, preprocessorId and debugMode

Tested on unicorn by running preprocessors both parallely and sequentailly.

@jeffbl @JRegimbal 

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/orchestrator/src/server-cache.ts
+++ b/orchestrator/src/server-cache.ts
@@ -25,13 +25,23 @@ export class ServerCache {
         this.memjsClient = Client.create();
     }
     async getResponseFromCache(hashedKey: string){
-        const cacheResponse = await this.memjsClient.get(hashedKey);
-        return cacheResponse && cacheResponse.value?.toString(); 
+        try{
+            const cacheResponse = await this.memjsClient.get(hashedKey);
+            return cacheResponse && cacheResponse.value?.toString(); 
+        } catch(error){
+            console.debug("Error getting response from the cache");
+            return undefined;
+        }
+
     }
     
     async setResponseInCache(hashedKey: string, value: string, timeout: number){
         console.debug(`storing data in memcache with key ${hashedKey}`);
-        await this.memjsClient.set(hashedKey, value, {expires: timeout});
+        try{
+            await this.memjsClient.set(hashedKey, value, {expires: timeout});
+        } catch(error){
+            console.debug("Error setting response in the cache");
+        }
     }
     
     constructCacheKey(data: Record<string, unknown>, preprocessor: string): string{


jeffbl:
On unicorn and pegasus at least, 4GB RAM use for cache shouldn't be an issue. Hopefully that buys us a fair number of requests cached in practice...

JRegimbal:
@jaydeepsingh25 could you add a bit more information on how you tested the *cache* in particular? It's unclear what was actually done based on the PR description.

orchestrator/src/server.ts
```diff
@@ -89,30 +95,64 @@ async function runPreprocessorsParallel(data: Record<string, unknown>, preproces
...
+                    }).then(r => {
+                        clearTimeout(timeout);
+                        const response = r.clone();
+                        if(response.status == 200){
+                            response.json().then((json) => {
+                                if (ajv.validate("https://image.a11y.mcgill.ca/preprocessor-response.schema.json", json)) {
+                                    SERVICE_PREPROCESSOR_MAP[preprocessor[0]] = json["name"];
+                                    // store data in cache
+                                    // disable the cache if "ca.mcgill.a11y.image.cacheTimeout" is 0
+                                    if(cacheTimeOut){
```
JRegimbal, line 136:
```suggestion
                                    if(cacheTimeOut > 0){
```
Personal issue with using implied conversions since they're a *pain* to debug.
For example, you're not casting "cacheTimeOut" to a number above explicitly, so `if(0)` would resolve to false, but `if("0")` would resolve to true. However, both `0 > 0` and `"0" > 0` are false.

orchestrator/src/server.ts
```diff
@@ -355,6 +424,17 @@ app.get("/authenticate/:uuid/:check", async (req, res) => {
     }
 });
 
+async function getResponseFromCache(hashedKey: string){
```
JRegimbal, line 427: Are these functions just meant to provide an abstraction around caching server calls?

docker-compose.yml
```diff
@@ -9,6 +9,7 @@ services:
     environment:
       - PARALLEL_PREPROCESSORS=ON
       - STORE_IMAGE_DATA=ON
+      - MEMCACHE_SERVERS=memcached
```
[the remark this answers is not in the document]
JRegimbal: This also doesn't appear to be used...

orchestrator/src/server.ts
```diff
@@ -30,6 +31,11 @@ import definitionsJSON from "./schemas/definitions.json";
 import { docker, getPreprocessorServices, getHandlerServices } from "./docker";
 
 const app = express();
+const memjsClient = Client.create();
```
JRegimbal, line 34: I don't like caching server selection being purely automatic. I can imagine users wanting to be able to have multiple servers running for different purposes, and having issues with how this is currently configured.

JRegimbal (changes requested):
Overall, the flow seems to match what we discussed: check if there's a timely entry saved in the cache, use that if available, and otherwise make the request and save the result if successful.

However, there seem to be some issues in how the caching is configured (or cannot be configured) and in some aspects of the implementation. Let me know if anything needs to be clarified.

orchestrator/src/server.ts
```diff
@@ -89,30 +95,64 @@ async function runPreprocessorsParallel(data: Record<string, unknown>, preproces
...
-            }).catch(err => {
-                console.error("Error occured fetching from " + preprocessor[0]);
-                console.error(err);
-                resolve();
+            // get value from cache for each preprocessor if it exists
+            const cacheTimeOut = preprocessor[3] as number;
+            const reqCapabilities = data["capabilities"] as string[];
+            const isDebugMode = reqCapabilities && reqCapabilities.includes("ca.mcgill.a11y.image.capability.DebugMode")
+            const preprocessorName = SERVICE_PREPROCESSOR_MAP[preprocessor[0]] || '';
+            const cacheKeyData = {"imageBlob": data["graphic"], "preprocessor":preprocessorName, "debugMode":isDebugMode};
```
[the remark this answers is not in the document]
jeffbl: Not ideal, but if you do this, please make sure to link to new work item here, to make sure it is on project board and won't get lost.

orchestrator/src/server.ts
```diff
@@ -89,30 +95,64 @@ async function runPreprocessorsParallel(data: Record<string, unknown>, preproces
...
+                                if (ajv.validate("https://image.a11y.mcgill.ca/preprocessor-response.schema.json", json)) {
+                                    SERVICE_PREPROCESSOR_MAP[preprocessor[0]] = json["name"];
+                                    // store data in cache
+                                    // disable the cache if "ca.mcgill.a11y.image.cacheTimeout" is 0
+                                    if(cacheTimeOut){
+                                        console.debug(`Saving Response for ${json["name"]} in cache with key ${hashedKey}`);
+                                        setResponseInCache(hashedKey, JSON.stringify(json["data"]), cacheTimeOut).then(()=>{
+                                            console.debug(`Saved Response for ${json["name"]} in cache with key ${hashedKey}`);
+                                        });
+                                    }
```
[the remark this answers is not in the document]
jeffbl: Yes, absolutely. But as much as possible should go in repo README and associated files. Wiki is more for higher-level documentation.

orchestrator/README.md
```diff
@@ -79,3 +80,22 @@ be omitted if the default behavior is desired.
...
+## Cache Implementation
+
+IMAGE uses Memcached as in-memory data store. Cache is implemented using [MemJS](https://www.npmjs.com/package/memjs). Following is the confugration to enable Cache for preprocessors:
+
+- Cache size is configured in the docker-compose in the commad attribute under memcached service `command: -m 4096` implies cache size of 4GB.
+
+- Cache timeout is configured at the preprocessor level, with the label `ca.mcgill.a11y.image.cacheTimeout` . Label value is the timeout value in seconds. Timeout value of 0 indicates that Cache is disabled for a preprocessor. Missing `ca.mcgill.a11y.image.cacheTimeout` label on the preprocessor will default to timeout value of 0.
+
+- Cache key is generated using the following attributes:
+  - `reqData` can have the following falues
```
JRegimbal, line 93:
```suggestion
  - `reqData` can have the following values
```

orchestrator/README.md
```diff
@@ -79,3 +80,22 @@ be omitted if the default behavior is desired.
...
+    - `request["data"]` (for graphics)
+    - `request["placeID"]`/`request["coordinates"]` (for maps)
+    - `request["highChartsData"]` (for charts)
+  - `preprocessor` - preprocessor id as returned in the response
```
JRegimbal, line 97: For clarity, this ID is **not** identifying the preprocessor itself, it is identifying the _data returned by the preprocessor_. This is why we have greated multiple object detection preprocessors that all use the same ID, they are separate components that produce data with the same semantic meaning and the same formatting.

orchestrator/README.md
```diff
@@ -79,3 +80,22 @@ be omitted if the default behavior is desired.
...
+  - `debugMode` - true/false depending on if the debug mode is enabled or not.
```
JRegimbal, line 98: Looking at this again, is there actually a case where debug mode _should_ impact the results of a preprocessor? I can't think of anything and am leaning towards this being removed. Thoughts, @jeffbl?

orchestrator/src/server-cache.ts
```diff
@@ -0,0 +1,36 @@
+import { Client } from "memjs";
```
JRegimbal, line 1: Could you include our AGPL header up here?

docker-compose.yml
```diff
@@ -9,6 +9,7 @@ services:
     environment:
       - PARALLEL_PREPROCESSORS=ON
       - STORE_IMAGE_DATA=ON
+      - MEMCACHE_SERVERS=memcached:11211
```
JRegimbal, line 12: Is this ever actually used in code? Under the hood by memjs?

orchestrator/src/server.ts
```diff
@@ -89,30 +95,64 @@ async function runPreprocessorsParallel(data: Record<string, unknown>, preproces
...
-                "signal": controller.signal
-            }).then(r => {
-                clearTimeout(timeout);
-                resolve(r);
-            }).catch(err => {
-                console.error("Error occured fetching from " + preprocessor[0]);
-                console.error(err);
-                resolve();
+            // get value from cache for each preprocessor if it exists
+            const cacheTimeOut = preprocessor[3] as number;
```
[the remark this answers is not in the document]
JRegimbal: (I'm suggesting you make sure that this can be meaningfully cast to a numeric type)

JRegimbal (changes requested):
Minor changes/questions, particularly on debug mode and the MEMCACHE_SERVERS environment variable.

orchestrator/README.md
jeffbl: Good question. The idea when we first discussed was that preprocessors (or eventually handlers, if we end up caching any of those in the future) might change their behavior based on the debug (or other capabilities) flags. But as @JRegimbal pointed out in conversation this morning, if the preprocessors are, say, outputting more log information when they see the debug flag, that won't appear past the first invocation anyway, if we pay attention to the debug flag. If preprocessors are changing their actual results, then it really isn't a good proxy for actual use, and it should really be some other flag other than "debug". I feel like there may be a valid reason to do this, but until we can articulate it, agree debug can be removed.

However, this also begs the question of other capabilities. I could argue it would be easiest to simply include all capabilities (including debug). Worst case, e.g., for debug, we do one extra uncached runthrough, to be sure that if someone does do something based on any capability, they don't get an incorrect cached version. At least this way the story is consistent - any capability difference causes a cache miss. Once we start picking and choosing, even for good reasons, it becomes more difficult for developers to know what to expect, if they do things we don't anticipate. @jaydeepsingh25 thoughts?

JRegimbal (approved):
Changes look good, I just want to check one thing before merge: when the MEMCACHE_SERVERS variable is unset, what behavior do we get @jaydeepsingh25?

Issue #830, opened 2024-05-29
Improve logging for preprocessor output


jaydeepsingh25:
Improve server logging for preprocessors so that the output is logged in the server logs. This should be enhanced for all preprocessors with relatively simple output.

PR #838, opened 2024-06-12
fix: add debug output

jaydeepsingh25:
Resolves https://github.com/Shared-Reality-Lab/IMAGE-server/issues/830 . Added logging for preprocessors with relatively simple outputs. Some of them with huge response output are not logged.

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/preprocessors/content-categoriser/Dockerfile
+++ b/preprocessors/content-categoriser/Dockerfile
@@ -8,7 +8,7 @@ RUN adduser --disabled-password python
 WORKDIR /app
 ENV PATH="/home/python/.local/bin:${PATH}"
 
-RUN pip3 install pip==24.0
+RUN pip3 install --upgrade pip
 RUN pip3 install -r /app/requirements.txt
 
 COPY /schemas /app/schemas
--- a/preprocessors/content-categoriser/requirements.txt
+++ b/preprocessors/content-categoriser/requirements.txt
@@ -1,9 +1,9 @@
 Flask==2.2.5
 flask_api==2.0.0
 numpy==1.21.0
-pillow==10.2.0
+pillow==10.3.0
 jsonschema==3.2.0
 Werkzeug==3.0.3
-pytorch_lightning==1.6.0
+pytorch_lightning==1.9.0
 opencv-python== 4.5.1.48
 gunicorn==22.0.0
--- a/preprocessors/grouping/grouping.py
+++ b/preprocessors/grouping/grouping.py
@@ -99,6 +99,7 @@ def readImage():
     for i in range(len(group)):
         for j in range(len(group[i])):
             dummy[i].append(group[i][j][0])
+        logging.debug("Number of Objects in group - " + len(dummy[i]))
         final_group.append({"IDs": dummy[i]})
 
     for i in range(len(check_group)):
@@ -107,6 +108,8 @@ def readImage():
     request_uuid = content["request_uuid"]
     timestamp = time.time()
     name = "ca.mcgill.a11y.image.preprocessor.grouping"
+    logging.debug("Number of groups ", len(final_group))
+    logging.debug("Number of ungrouped objects ", len(ungrouped))
     data = {"grouped": final_group, "ungrouped": ungrouped}
     try:
         validator = jsonschema.Draft7Validator(data_schema)
--- a/preprocessors/yolov8/detect.py
+++ b/preprocessors/yolov8/detect.py
@@ -186,6 +186,7 @@ def detect_objects(send,
                         "confidence": np.float64(label[-4:]),
                         "centroid": centre, "area": area
                     }
+                    logging.debug("Object Detected - " + dictionary.type)
                     send.append(dictionary)
                     """"plot_one_box(xyxy, im0, label=label,
                     color=colors(c, True),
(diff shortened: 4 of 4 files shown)


preprocessors/caption-recognition/caption.py
```diff
@@ -87,7 +87,8 @@ def captions():
     except jsonschema.exceptions.ValidationError as e:
         logging.error(e)
         return jsonify("Invalid Preprocessor JSON format"), 500
-    logging.debug(data)
+    logging.debug("Caption Recoginition Response")
```
jeffbl, line 90:
```suggestion
    logging.debug("Caption Recognition Response")
```

jeffbl: For preprocessors that output information tightly related to the content, we have to be very careful what we log, for privacy reasons. It is fine to put things like categorizing content as a photo or chart, for example, but something like OCR would be a big problem, since we do not want to retain that information. Unless we're sure that the content of the response will not be a privacy issue, we need to remove this, or output something more abstract, but still potentially informative.

preprocessors/celebrity-detector/celebrity-detector.py
```diff
@@ -214,7 +214,8 @@ def categorise():
         except jsonschema.exceptions.ValidationError as e:
             logging.error(e)
             return jsonify("Invalid Preprocessor JSON format"), 500
-        logging.debug("Sending response")
+        logging.debug("Celebrity Detector Response")
+        logging.debug(response)
```
jeffbl, line 218: again potential privacy issue?

preprocessors/collage-detector/detect.py
```diff
@@ -91,7 +91,8 @@ def detect_collage():
     except jsonschema.exceptions.ValidationError as e:
         logging.error(e)
         return jsonify("Invalid Preprocessor JSON format"), 500
-    logging.debug("Sending response")
+    logging.debug("Collage Detector Response")
+    logging.debug(response)
```
jeffbl, line 95: If this will output on two separate lines, should we combine into one to allow for easier grep, etc.?

jeffbl (changes requested):
Big issues are:
1. dumping full response is likely to be very large in the logs?
2. dumping full response may create privacy issues, depending on the preprocessor
3. splitting log output on multiple lines will make it hard to do things like grep

Suggest that we do more specific output for each preprocessor. For example, for collage detector, just a single line e.g., "collage-preprocessor: is a collage" or "collage-detector: not a collage".

jeffbl:
Assigning to @JRegimbal for further comments/direction before returning to @jaydeepsingh25 for implementation.

preprocessors/caption-recognition/caption.py
JRegimbal, line 90: If you're interested in having a clear indication of the preprocessor on each line (beyond what docker will do anyway in compose mode), you can get this easily with [a named logger](https://docs.python.org/3/library/logging.html#logging.getLogger). It would only require making calls through the logger object rather than the `logging` module, although it may not be crucial for smaller preprocessors like this.

preprocessors/celebrity-detector/celebrity-detector.py
JRegimbal: Possibly...maybe something like "detected [n] celebrities out of [m] objects"?

preprocessors/collage-detector/detect.py
JRegimbal: Addressed this a bit in an earlier comment, but using a named logger could help. Docker compose does give some naming in the overall compose logs. Including some below:
```
content-categoriser-1          | [2024-06-19 11:39:05 +0000] [9] [DEBUG] POST /preprocessor
orchestrator-1                 | Sending to preprocessor "object-sorting"
```

JRegimbal: (To be clear, I think this is likely sufficient for our purposes?)

JRegimbal (dismissed):
Like Jeff, I'm skeptical that logging the entire response is the right way to approach this. A summary more readable to a person at a higher level than the raw data is probably the way to go here.

On the point of including the name of the preprocessor in each line/message, I think the service name we get in the docker logs is probably sufficient for all of the services here. If anything gets bigger and has multiple modules working together, then I think using a named logger might make the logs a bit clearer.

jeffbl (approved):
Looks ok, but I see you removed any output for preprocessors like yolo and grouping. These could at least give a count if that is easy to do. Those can be added in a separate PR if they make sense.

preprocessors/content-categoriser/Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN adduser --disabled-password python
 WORKDIR /app
 ENV PATH="/home/python/.local/bin:${PATH}"
 
-RUN pip3 install --upgrade pip
+RUN pip3 install pip==24.0
```
JRegimbal, line 11: Any particular reason for this change?

JRegimbal (commented):
I'm skeptical of removing the output of processors that produce the most data, since these would be the hardest to debug. Maybe we discuss this during the meeting?

PR #865, opened 2024-08-04
fix: deploy high-charts-preprocessor

### LINKED ISSUE #857: Highcharts preprocessor is not deployed
Earlier, we had created a Highcharts preprocessor to allow the creation of synthetic data columns to help out the handlers. This was never deployed, and therefore cannot be cached. Work item is to make sure the preprocessor is working correctly and update the core docker-compose to enable it in test, then production.

jaydeepsingh25:
resolves https://github.com/Shared-Reality-Lab/IMAGE-server/issues/857

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/preprocessors/line-charts/charts.py
+++ b/preprocessors/line-charts/charts.py
@@ -25,6 +25,7 @@ from charts_utils import getLowerPointsOnLeft, getHigherPointsOnLeft
 from charts_utils import getLowerPointsOnRight, getHigherPointsOnRight
 
 app = Flask(__name__)
+logging.basicConfig(level=logging.DEBUG)
 
 
 @app.route('/preprocessor', methods=['POST', 'GET'])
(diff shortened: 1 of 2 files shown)


preprocessors/line-charts/charts.py
```diff
@@ -124,7 +125,7 @@ def get_chart_info():
         return jsonify("Invalid Preprocessor JSON format"), 500
 
     logging.debug("Sending response")
-    # print(data)
+    logging.debug(data)
```
jeffbl, line 128: @jaydeepsingh25 Not sure we want to dump all of the actual data into logs.

jeffbl: I tested with the [etherscan.io chart that contains many points](https://etherscan.io/chart/etherprice), and the single log line from line-chart is 438K alone, amking it very difficult to watch the overall logs:

`-rw-rw-r-- 1 jeffbl jeffbl 438K Aug  7 14:53 /home/jeffbl/foo.log`

Plus if it outputs any of the title, axes, etc. then this is end-user content we should not be storing in logs.

Can we just summarize it with something like the number of points, total size, or some other proxy value to give an idea of the response?

PR #901, opened 2024-10-19
Return valid osm response for placeId

jaydeepsingh25:
The PR provides fix for https://github.com/Shared-Reality-Lab/IMAGE-server/issues/870.
It fixes a common bug for the cases where map request is made with a query text, OSM preprocessor does not return a valid response in that case. 
PlaceId in the request in the request contains the queryText and not actual placeId. Hence we need to call `place/textsearch `endpoint rather than `place/details` endpoint. Autour preprocess has a similar logic in place.
Tested on unicorn with both the cases(coordinates and placeId)

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/preprocessors/ner/requirements.txt
+++ b/preprocessors/ner/requirements.txt
@@ -7,7 +7,7 @@ torch==1.13.1
 tqdm~=4.66.3
 torchvision==0.8.2
 clip-by-openai==1.1
-scikit-learn==1.5.0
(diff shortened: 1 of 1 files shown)


jeffbl (approved):
looks like it just brings OSM in line with what Autour was already doing. Looks good!

Issue #939, opened 2024-12-24
Provide caption-only for content-categoriser non-photo results (reduce "IMAGE is unable to render this graphic" errors) 


jeffbl:
The [recently deployed LLM-based content-categoriser](https://github.com/Shared-Reality-Lab/IMAGE-server/pull/914) should be better about identifying photos as photos. However, there are still plenty of graphics on the web that will be (correctly) categorised as [other](https://kagi.com/proxy/cartoon-pokemon-style-monster-from-fotor-ai-pokemon-generator_2023-03-30-103225_paeg.jpg?c=3WTGQYZC-74NdqOwGOEGFgggu-flrBVuoBpaIf_E9N2To1Ytc-ZO7Hh6o9tYZK3C2zOfnL3BLPpy_R8wcZJxnLuLMVcMlZCSr2tq3vCm4BCgfyGxDrcPlpiEdJzG121n2ICrrIMTW6f88-jB7Tx_1ZgVye52nSqn3DiWrp5W95v0vlSegjihqVkCeypDHcKERjvplS3nrNyTmjN3x964SA%3D%3D), [text](https://onlinejpgtools.com/images/examples-onlinejpgtools/book-cover.jpg), or [chart](https://kagi.com/proxy/bar-1.gif?c=NhrQ1VJ9jwxpTXDw3W-sx61f8deziBcCWwYMqlnPTR0sd0V2fRpFgkE6wTnX5ZHyb-VtluB8VuJkgiTmIeeG4Nk37wHSc6EOmHWjzteChZPaVqpMWDkPkyzqTlxMlj6Z). Right now, these simply return an error saying IMAGE is unable to render the graphic. This is confusing, and irritating since it can happen frequently depending on what you're looking at. Right now, photo-audio-handler ignores everything not categorised "photo" (hence the name of the handler!) but this then just errors out since we don't have handlers for these other types.

REJECTED PROPOSAL: Extend photo-audio-handler to deal with graphics other than photos, and just provide the caption if there is nothing else that can be obtained. I see this as a worse option since most of the photo-audio-handler is focused on photo-specific tasks such as extracting objects and regions, that likely don't apply to non-photo graphics, and could go off the rails producing spurious or wrong information.

[...]

PR #941, opened 2025-01-13
Initial commit using captioner

### LINKED ISSUE #933: STORY: As an audio-only IMAGE user, I want an overall scene description, so that the information I get about specific items via spatialized audio makes sense in context.
We created the graphic-caption preprocessor primarily for the Monarch experiences. However, participants have mentioned the desire to have more of an overview in the audio-only experiences as well. This work item is to integrate the output of `graphic-caption` preprocessor at the beginning of the photo-audio-handler experiences, to give the user a clear context for the specific information about objects and such that they receive after.

Note that very early in the project we did have a captioner, but deprecated it since it  was giving biased results. Hopefully llama3.2-vision (current LLM in use) has better protections than what we were using in 2021.

@Cybernide Please provide instructions on where it should appear in the experience. I assume at the beginning, but I don't know if there are tweaks to the overview, or other introductory text that should introduce the description. Once you've put those instructions in this work item, please assign to @JRegimbal. 

@JRegimbal We can revisit who does this when you receive it, if you don't have IMAGE hours to spend, or this turns out to be more involved than I'm assuming. (Which is that it should be a straightforward addition, especially if any of the old code is still there from our original captioner)

JRegimbal:
This resolves #933 by checking for preprocessor data under the `graphic-caption` object. If a caption is present, it is inserted at the start of the description, before semantic segmentation or object detection. Changes are made so that if only the caption information is available, the handler will still run.

There is likely to be considerable redundancy between these three sources of data.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/handlers/photo-audio-handler/src/server.ts
+++ b/handlers/photo-audio-handler/src/server.ts
@@ -62,11 +62,12 @@ app.post("/handler", async (req, res) => {
     const objDet = preprocessors["ca.mcgill.a11y.image.preprocessor.objectDetection"];
     const objGroup = preprocessors["ca.mcgill.a11y.image.preprocessor.grouping"];
     const action = preprocessors["ca.mcgill.a11y.image.preprocessor.actionRecognition"];
-    const collageDetector = preprocessors["ca.mcgill.a11y.image.preprocessor.collageDetector"]
+    const collageDetector = preprocessors["ca.mcgill.a11y.image.preprocessor.collageDetector"];
+    const graphicCaption = preprocessors["ca.mcgill.a11y.image.preprocessor.graphic-caption"];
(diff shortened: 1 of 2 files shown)


jeffbl:
I don't see any red flags from looking at the code, but I also don't see where a change was made to ignore content-categoriser, which I thought it paid attention to in the past? Also, in terms of redundancy, if someone later makes something like a `textgraphic-audio-handler` that does a nicer job but also puts in the caption, then this will still put out a caption-only rendering, which will be completely redundant? Note the proposal in #939, which I don't see implemented? If you have time, can we sync up in person on this so that I understand the implications, and if needed, we can log any new work items before this gets merged?

Issue #903, opened 2024-10-30
Zero length audio file returned for map request, and debug view is messed up.


jeffbl:
Follow steps in #871 
When it is finding the place, the returned experience is zero length, and the debug information is messed up:

![image](https://github.com/user-attachments/assets/89d131ad-aca6-44d8-a42d-ac4c437f590d)

Expect:
- if no valid response can be generated (and an empty audio file should not be valid) no experience is returned at all
- expanding debug area should not draw over other experience areas in window (let me know if should break out into separate issue)

JRegimbal:
Hi all, I'm unable to reproduce the zero-length audio file issue after several attempts. I'm receiving an audio representation of streets as expected. Can someone else try to replicate?

jeffbl:
just tried against unicorn, and also get a valid non-zero length rendering
so it was most likely some other problem, like the autour preprocessor not running/failing? Is it worth a quick code inspection to see if there is an obvious path that would return empty sound renderings?

JRegimbal:
Closing since Jeff wants it to disappear (will reopen if the issue comes back).

PR #963, opened 2025-02-17
Filter autour places before checking length

### LINKED ISSUE #962: Autour handler checks number of places before filtering
While the autour handler does check that there is at least one place listed on OSM around the point-of-interest, this check is performed _before_ filtering out paths.

https://github.com/Shared-Reality-Lab/IMAGE-server/blob/56894689fdea2b177c9778b1539c245adf393b63/handlers/autour-handler/src/server.ts#L58

https://github.com/Shared-Reality-Lab/IMAGE-server/blob/56894689fdea2b177c9778b1539c245adf393b63/handlers/autour-handler/src/server.ts#L94

If the places object only contains data that will be filtered (currently, only ways), then it is possible that the resulting rendering will contain no information.

This may be the cause behind #903, however this cannot be confirmed since the issue cannot be replicated at this time.

JRegimbal:
This resolves #962 by filtering the autour places returned by the preprocessor by category before checking that at least one place is available. Previously, it was possible that if the places array only consisted of excluded categories, that an audio file would be generated anyway, perhaps leading to zero-length files such as #903.

This was tested with the map referenced in #871. The same rendering was returned with the updated handler.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/handlers/autour-handler/src/server.ts
+++ b/handlers/autour-handler/src/server.ts
@@ -55,7 +55,10 @@ app.post("/handler", async (req, res) => {
     }
 
     const autourData = preprocessors["ca.mcgill.a11y.image.preprocessor.autour"];
-    if (autourData["places"].length === 0) {
+
+    // Filter places we will not use before checking for non-zero length.
+    const places = autourData["places"].filter((p: { "cat": number }) => !filterCategories.includes(p["cat"]));
+    if (places.length === 0) {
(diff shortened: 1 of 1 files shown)


jeffbl:
No red flags from code inspection.

jeffbl:
(and did `./restoreunstable` on unicorn post-merge and verified that demo map on IMAGE website still produces renderings.)

PR #972, opened 2025-03-04
Add chat history preservation, deleteion, and cleanup

### LINKED ISSUE #970: text-followup preprocessor: add history for maintaining chat context
Right now, text-followup allows the user to ask a single question on the Monarch (e.g., "Is there a cat in the picture?"), and the LLM will send a text response that the Monarch then reads out (e.g., "Yes, there is a cat"). However, if the user asks a second question (e.g., "What color is it"), there is no history maintained, so the LLM will not provide a useful response. There are two possible approaches to this:

1. change from the `generate` open-webui endpoint to `chat`. @JRegimbal has done this for another project, so may have advice on any gotchas that will arise.
2. The history of the chat is also stored in the followup request, so could be resubmitted to the LLM as part of the prompt.

I'm not sure which approach is better, but the first one seems more straightforward if it gives us what we want. I also don't understand how long the server would cache this information, or what issues might arise there in a long-running conversation.

gvzdv:
Resolve #970 

Summary of changes:
* Set up in-memory storage for chat history
* Chats are stored based on the request_uuid
* Added chat history length and history expiry. TO DO: discuss these figures and add them to .env
* Changed API endpoint from `generate` to `chat` to be able to pass chat history. TO DO: change it in the .env file
* Modified ollama request structure to include chat history
* Modified ollama request structure to include JSON schema
* Added chat history logging. TO DO: move it under the `log_pii` condition
* Added chat history cleanup after set time
* Added an endpoint to delete chat history for a specific uuid. TO DO: discuss if it's needed
* Added an endpoint to get conversation statistics. TO DO: discuss if it's needed

Testing:
* Tested using jQuery requests and asking follow-up questions on the same picture
* Logging shows that the chat history is passed along with each new request from the same uuid
* Model answers are cohesive; the model is able to reference its previous responses
* Chat history cleanup works as intended: after the set time, the conversations are deleted, and new dialogues start from scratch - checked via logs.

To do:
* Items described in the summary
* Testing of `/clear-history` and `history-stats` endpoints

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/preprocessors/text-followup/text-followup.py
+++ b/preprocessors/text-followup/text-followup.py
@@ -26,6 +26,33 @@ from datetime import datetime
 app = Flask(__name__)
 logging.basicConfig(level=logging.DEBUG)
 
+# ADDED: Dictionary to store conversation history by request_uuid
+# We'll use this to maintain conversation context between requests
+conversation_history = {}
+
+# ADDED: Configuration for history management
+MAX_HISTORY_LENGTH = int(os.getenv('MAX_HISTORY_LENGTH', '50'))  # Max messages to keep
+HISTORY_EXPIRY = int(os.getenv('HISTORY_EXPIRY', '30'))  # History expiry in seconds
+
+
+# ADDED: Function to clean up old conversation histories
+def cleanup_old_histories():
+    """
+    Remove conversation histories that are older than HISTORY_EXPIRY seconds
+    """
+    current_time = time.time()
+    uuids_to_remove = []
+    
+    for uuid, history in conversation_history.items():
+        if current_time - history.get('last_updated', 0) > HISTORY_EXPIRY:
+            uuids_to_remove.append(uuid)
+    
+    for uuid in uuids_to_remove:
+        del conversation_history[uuid]
+    
+    if uuids_to_remove:
+        logging.debug(f"Cleaned up {len(uuids_to_remove)} old conversation histories")
+
 
 @app.route("/preprocessor", methods=['POST', ])
 def followup():
@@ -66,6 +93,15 @@ def followup():
     request_uuid = content["request_uuid"]
     timestamp = time.time()
     name = "ca.mcgill.a11y.image.preprocessor.text-followup"
+
+    # ADDED: Clean up history based on defined expiry time
+    cleanup_old_histories()
+
+    # ADDED: History status log
+    logging.debug(f"Current history status: {len(conversation_history)} conversations stored")
+    logging.debug(f"Request UUID {request_uuid} exists in history: {request_uuid in conversation_history}")
+    if request_uuid in conversation_history:
+        logging.debug(f"Current history length for {request_uuid}: {len(conversation_history[request_uuid]['messages'])}")
 
     # debugging this preprocessor is really difficult without seeing what
     # ollama is returning, but this can contain PII. Until we have a safe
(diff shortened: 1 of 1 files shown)


preprocessors/text-followup/text-followup.py
```diff
@@ -130,16 +167,70 @@ def followup():
...
     else:
         logging.warning("OLLAMA_API_KEY does not start with sk-")
+    
+    # MODIFIED: Create messages list for chat endpoint instead of single prompt
+    system_message = {
+        "role": "system",
+        "content": general_prompt
+    }
+    
+    # Discuss if we want to attach the image every time
```
jeffbl, line 177:
```suggestion
    # TODO: Discuss if we want to attach the image every time
```

preprocessors/text-followup/text-followup.py
```diff
@@ -214,6 +320,44 @@ def followup():
     return response
 
 
+# ADDED: New endpoint to clear history for a specific request_uuid
+@app.route("/clear-history/<request_uuid>", methods=["GET"])
+def clear_history(request_uuid):
```
jeffbl, line 325: Need to remember to put this on a cron job?

preprocessors/text-followup/text-followup.py
gvzdv: Based on the discussion with @JRegimbal, I'll try attaching the image once at the first request and track how it impacts the results in the long run.

preprocessors/text-followup/text-followup.py
```diff
@@ -130,16 +167,70 @@ def followup():
...
+    # ADDED: Create log-friendly version of messages without full base64 content
+    log_friendly_messages = []
+    for msg in messages:
+        log_msg = msg.copy()  # Create a copy to avoid modifying the original
+        if 'images' in log_msg:
+            # Replace image content with placeholder or truncate it
+            log_msg['images'] = [f"[BASE64_IMAGE:{len(img)} bytes]" for img in log_msg['images']]
+        log_friendly_messages.append(log_msg)
+
+    logging.debug(f"Message history: {log_friendly_messages}")
```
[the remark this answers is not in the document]
gvzdv: Yes, will move it under the if `log_pii` condition.

preprocessors/text-followup/text-followup.py
```diff
@@ -130,16 +167,70 @@ def followup():
...
+            "type": "object",
+            "properties": {
+                "response_brief": {
+                    "type": "string"
+                },
+                "response_full": {
+                    "type": "string"
+                }
+                },
+                "required": [
```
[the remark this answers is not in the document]
gvzdv: No, it just reiterated the schema that the model should produce. There's an error in the `required` parameters—I know about it and will fix it in the next commit.

preprocessors/text-followup/text-followup.py
gvzdv: I changed the code to include the image once (at the first request) and didn't see any performance drop after testing: the image is kept in the conversation history and the model is able to parse additional details from it on request. If we notice significant problems with that later, it is easy to revert this change.

preprocessors/text-followup/text-followup.py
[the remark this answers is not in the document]
gvzdv: Done. Moved the entire process of conversation history log creation under the `log_pii` condition to skip it altogether in production.

preprocessors/text-followup/text-followup.py
gvzdv: As I mentioned, the preprocessor automatically clears all conversations older than 1 hour when it's called.
If we want to delete outdated conversations even when the system is not in use, we can put it on a cron job via the `GET` request. I converted the `cleanup_old_histories()` function into a `/cleanup` route.

preprocessors/text-followup/text-followup.py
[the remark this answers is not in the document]
gvzdv: To clarify: this schema ensures that the model output follows the schema described in the prompt and includes both `response_brief` and `response_full`.

gvzdv:
@jaydeepsingh25 all comments from @jeffbl are addressed. Let me know if you have any concerns.
To properly work, this should be merged alongside #976. I have included the instructions for changing the config file in the comments to that PR but decided to keep them separate for traceability.

gvzdv:
@jaydeepsingh25, sorry about that! Conflicts resolved, should be good to go now.

Issue #982, opened 2025-03-13
Fix text followup preprocessor caching


jaydeepsingh25:
Text followup preprocessor fails to give correct output if caching is turned on. 

Current behavior: 
(Enable cache for text-followup preprocessor - `ca.mcgill.a11y.image.cacheTimeout`: 3600)
 Preprocessor gives same response for a given image even if the followup-query is different because response is served from cache.

![Image](https://github.com/user-attachments/assets/6f1bd5a9-e427-4882-969a-3f2fd0ed6830)

Expected Behaviour: 
For a given image, the preprocessor response should be different if followup query is different. 

Potential fix is to include followup-query in hashkey. Currently [hashkey ](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/orchestrator/src/server-cache.ts#L60) just takes into account preprocessor name and graphic

jeffbl:
Note that if not trivial for any reason, temporary solution is to simply turn off caching in the docker-compose for the text-followup preprocessor.

In solving this, we also need to think about a way to allow other preprocessor creators to specify additional fields that need to be examined for caching, since this is unlikely to be the only time a preprocessor has some unique need for additional fields to be part of the hash.

PR #983, opened 2025-03-14
fix: Include followup query in cache key

jaydeepsingh25:
The PR provides fix for issue: https://github.com/Shared-Reality-Lab/IMAGE-server/issues/982. 
The fix is to include followup query in the hashkey. Without this, stale data was wrongly served from the cache. After the fix, when the cache is enabled the preprocessor gives different responses for the same graphic (depending on followup query).

Before Fix:
image

After Fix:
image

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/orchestrator/src/server-cache.ts
+++ b/orchestrator/src/server-cache.ts
@@ -18,6 +18,12 @@
 import { Client } from "memjs";
 import hash from "object-hash";
 
+interface CacheValue {
+    reqData: string | object;
+    preprocessor: string; 
+    followupQuery?: string 
+}
+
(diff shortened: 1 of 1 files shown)


jeffbl (approved):
Technically no issues with doing this to make it work, but please see individual comment regarding making this generic going forward (which should be captured in a new work item). Sending to Mike for additional technical review.

jeffbl:
@gvzdv for technical review before merge

gvzdv (approved):
No concerns from my side regarding this fix as a temporary but effective measure.
I agree with Jeff on the need to implement a more robust and generalized way to track this data, probably by including the list of the values requiring caching in the preprocessor output JSON schema.

PR #984, opened 2025-03-16
Integration of Graph into image execution

### LINKED ISSUE #958: Retrieving and executing preprocessors and handlers through a dependency-mapped graph.

M-A89:
resolves #958
 
This PR allows for generating a dependency graph to map image preprocessor dependencies. It takes into consideration both optional and required dependencies which are retrieved through the docker-compose override for now. 
As this is still in the testing stage, I have kept the old "runPreprocessorsParallel" and added a "runPreprocessorsParallelGraph" and after testing we can utilize it for series & parallel execution. 
Please note that to test this PR you will need to include two new tags in the "Labels" section of the docker compose override and for the next sprint they will be added to the docker-compose.

---

## Required Information

## Coding/Commit Requirements

OR


JRegimbal:
Is this ready for me to review, @M-A89?

M-A89:
> Is this ready for me to review, @M-A89?

Yep should be all good

orchestrator/src/graph.ts
```diff
@@ -0,0 +1,128 @@
...
+          if (requiredPreprocessors.length === 0 && tmp.length === 0) {
+            R.add(node);
+          } else {
+            for(const reqPrep of requiredPreprocessors){
+              const nodePrime = this.addNode(reqPrep);
+              node.parents.add(nodePrime);
+              nodePrime.children.add(node);
+            }
+
+            for(const optPrep of optionalPreprocessors){
```
JRegimbal, line 44: There may be problems with how `optionalPreprocessors` is being constructed, but this should be `tmp` since we only want to iterate on optional preprocessors that are in the set of available preprocessors.

orchestrator/src/graph.ts
```diff
@@ -0,0 +1,128 @@
...
+              const nodePrime = this.addNode(optPrep);
```
JRegimbal, line 45: Will this not end up creating duplicate nodes for certain preprocessors?

orchestrator/src/graph.ts
```diff
@@ -0,0 +1,128 @@
...
+          }
+        }
+      }
+      //should be false if cycle exists
+      return processedCount === this.nodes.size;  
+    }
+}
+  
+export class GraphNode {
+    preprocessor: (string | number)[];    //preprocessor that this node represents
```
JRegimbal, line 92: Nodes are also handlers!

orchestrator/src/server.ts
```diff
@@ -30,6 +30,9 @@ import responseSchemaJSON from "./schemas/response.schema.json";
 import definitionsJSON from "./schemas/definitions.json";
 import { docker, getPreprocessorServices, getHandlerServices, DEFAULT_ROUTE_NAME } from "./docker";
 import { ServerCache } from "./server-cache";
+import { Graph } from "./graph";
+import { GraphNode } from "./graph";
+import { printGraph } from "./graph";
```
JRegimbal, line 35:
```suggestion
import { Graph, GraphNode, printGraph } from "./graph";
```

orchestrator/src/server.ts
```diff
@@ -162,45 +165,43 @@ async function executePreprocessor(preprocessor: (string | number)[], data: Reco
...
-
+    //let currentPriorityGroup: number | undefined = undefined;
+    //const queue: (string | number)[][] = []; //Microservice queue for preprocessors and handlers
+    const prepQueue = Array.from(R);    //queue of preprocessors 
 
     //function that dequeues everything in the queue at once, executes them and waits for them to finish processing
     const processQueue = async (): Promise<void> => {
         try {
-            await Promise.all(queue.map(preprocessor => executePreprocessor(preprocessor, data)));
+            await Promise.all(prepQueue.map(preprocessor => executePreprocessor(preprocessor.preprocessor, data)));
```
JRegimbal, line 179: Ideally we would not need to wait for every element of the queue to finish before we dispatch new requests.

orchestrator/src/server.ts
```diff
@@ -162,45 +165,43 @@ async function executePreprocessor(preprocessor: (string | number)[], data: Reco
...
+                        prepQueue.push(child);  
+                    }
+                }
+            }
         } catch (error) {
-            console.error(`One or more of the promises failed at priority group ${currentPriorityGroup}.`, error);
-        }
-        finally {   //empty the queue 
-            queue.length = 0;
+            console.error(`One or more of the promises failed at priority group.`, error);
```
JRegimbal, line 194: No longer priority groups.

orchestrator/src/server.ts
```diff
@@ -310,10 +351,23 @@ app.post("/render", (req, res) => {
             const preprocessors = getPreprocessorServices(containers, route);
             const handlers = getHandlerServices(containers, route);
 
+            const graph = new Graph();
+            //Construct the graph using the handlers and preprocessors 
+            const readyToRun =  await graph.constructGraph(preprocessors, []);
```
JRegimbal, line 356: Graph should also include handlers.

JRegimbal (changes requested):
Great work so far! I have a few notes/changes. Could you also provide more information on the following:

* What kind of testing have you done? How have you looked for edge cases, etc.
* Do you plan on including handlers in the graph at a future stage? As of right now, everything is written as if only preprocessors are part of the graph.

M-A89:
> Great work so far! I have a few notes/changes. Could you also provide more information on the following:
> 
> * What kind of testing have you done? How have you looked for edge cases, etc.
> * Do you plan on including handlers in the graph at a future stage? As of right now, everything is written as if only preprocessors are part of the graph.

Thank you Juliette for all your suggestions that I overlooked, I'll continue to apply them.  So far my testing has consisted of the following: 
- Running the orchestrator container without the Docker Compose override, saving the logs, and then comparing them with the logs from the override by comparing execution_time_ms and cpu_time_ms.
- Using the extension on the same image, ensuring that the old and new code have the same "not applicable" and "applicable" preprocessors.
- Testing for cyclic dependencies.
- Changing the optional and required dependencies in the override (so that they no longer match what the preprocessors actually need), which causes the logs to show unmet requirements, indicating that the preprocessors no longer work.
- Ensuring that the order of preprocessor execution matches what’s expected from the graph.
Apologies for overlooking the handlers! I’ve sent a more detailed explanation on Slack to keep this message short.

orchestrator/src/server.ts
M-A89: Hmm problem is Im not sure theres a way to directly know the exact minute a preprocessor is done other than by using await which would defeat the whole purpose.

orchestrator/src/docker.ts
```diff
@@ -101,3 +104,87 @@ export function getHandlerServices(containers: Docker.ContainerInfo[], route: st
...
+        const optionalPreprocessors = container.Labels[_OPTIONAL_DEP_LABEL_];
+        let optionalPreprocessorsArray : string[] = [];
+        
+        //convert from string to array of strings 
+        if(optionalPreprocessors != ""){    //if the preprocessor has dependencies
+            optionalPreprocessorsArray = optionalPreprocessors ? optionalPreprocessors.split(",").filter(Boolean) : [];
+        }
+        
+        //filter through the array of preprocessor names and return an array of their actual values
+        return preprocessors.filter(function (p) { return optionalPreprocessorsArray.includes(p[0] as string); });
```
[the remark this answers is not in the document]
M-A89: Since we get the optional and required preprocessor as we're building the graph, way before anything runs, the dependencies at this point are not running yet. (Sorry I could be misunderstanding your suggestion)

orchestrator/src/graph.ts
```diff
@@ -24,25 +29,31 @@ export class Graph {
...
 
-        if(optionalPreprocessors && requiredPreprocessors && requiredPreprocessors.every((r) => Pset.has(r[0] as string))){
-          const node = this.addNode(preprocessor);
+        if(optionalPreprocessors && requiredServices && requiredServices.every((r) => Pset.has(r[0] as string))){
+          const node = this.addNode(service, P.some(p => p[0] == service[0]));
           const tmp = optionalPreprocessors.filter((o) => Array.isArray(o) && Pset.has(o[0] as string));
 
-          if (requiredPreprocessors.length === 0 && tmp.length === 0) {
+          if (requiredServices.length === 0 && tmp.length === 0) {
             R.add(node);
```
JRegimbal: I think an explanatory comment here would help with readability!

orchestrator/src/graph.ts
JRegimbal: Still think this is relevant.

orchestrator/src/server.ts
```diff
@@ -163,68 +163,113 @@ async function executePreprocessor(preprocessor: (string | number)[], data: Reco
...
+async function runServicesParallel(data: Record<string, unknown>, preprocessors: (string | number)[][], G: Graph, R: Set<GraphNode>): Promise<Record<string, unknown>> {
     if (data["preprocessors"] === undefined) {
         data["preprocessors"] = {};
     }
-    const prepQueue = Array.from(R);  //Queue of preprocessors and handlers that are ready to run
+    
+    // Get unique set of nodes that are running
+    const running = Array.from(R)
+    .filter((service) => preprocessors.some(p => p[0] == service.value[0]))  // optional
+    .map((preprocesor) => executeGraphNode(preprocesor, data));
```
JRegimbal:
```suggestion
        .filter((service) => preprocessors.some(p => p[0] == service.value[0]))  // optional
        .map((preprocesor) => executeGraphNode(preprocesor, data));
```

JRegimbal (changes requested):
Looking much better! Only some minor suggestions at this point (keeping in mind plans for future commits). Could you give an update on how you're testing as well?

orchestrator/src/graph.ts
```diff
@@ -24,25 +29,31 @@ export class Graph {
...
-      for (const preprocessor of combinedArray) {
-        const requiredPreprocessors = await getRequired(preprocessor[0] as string, P);
-        const optionalPreprocessors = await getOptional(preprocessor[0] as string, P);
+      for (const service of combinedArray) {
+        const requiredServices = await getRequired(service[0] as string, P);
+        let optionalPreprocessors = [] as (string | number)[][];
+        //Only get the optional preprocessors if its a preprocessor
+        if(P.some(p => p[0] == service[0])){
+          optionalPreprocessors = await getOptional(service[0] as string, P);
+        }
```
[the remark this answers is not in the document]
M-A89: @JRegimbal since handlers will always require a preprocessor data, I assumed that every handler dependency will be required and that they don't have optional requirements, please feel free to correct me not too sure about this

shahdyousefak:
We still have the filtering logic left for getOptional() and getRequired(), which at the moment the containers list is an unfiltered pool of containers. As explained with Mira, not only is this including containers that may not be reachable within the scope of the orchestrator, but actually making redundant api calls may be impacting the performance too, as well as accruing a cost (?).

orchestrator/src/server.ts
```diff
@@ -311,26 +342,48 @@ function getRoute(data: Record<string, any>): string {
...
         // get route variable or set to default
-        let data = JSON.parse(JSON.stringify(req.body));
+        let data = JSON.parse(JSON.stringify(requestBody));
         const route = getRoute(data);
         // get list of preprocessors and handlers
         docker.listContainers().then(async (containers) => {
-            const preprocessors = getPreprocessorServices(containers, route);
-            const handlers = getHandlerServices(containers, route);
+            //Get the list of filtered containers that are connected to one of the Orchestrator networks
+            const connectedContainers = getFilteredContainers(containers);
```
shahdyousefak, line 358: const connectedContainers = getFilteredContainers(containers).filter(c => c.State === "running");

-- we need to check for running containers

shahdyousefak (changes requested):

orchestrator/src/server.ts
shahdyousefak: or better, put this inside getFilteredContainers()

shahdyousefak (approved):

PR #986, opened 2025-03-17
Refactor Object Detection preprocessor and update to YOLOv11

### LINKED ISSUE #910: update yolo v8 to latest version (currently 11)
Yolo v11 looks like it is more efficient and accurate:

https://docs.ultralytics.com/models/yolo11/

Task is to update to this version, maintaining the same schema. Assuming it is mostly the same in terms of the currently yolov8 preprocessor, can just make the update in the current one, but rename to yolo (to be version agnostic for the preprocessor name), and give version and other info in the README.

Note this may impact #247, since different versions may do better/worse with different graphic input sizes. (In other words, I'd do the update first, then optimize graphic dimensions)

gvzdv:
Fixed #910.

- Refactored the object detection preprocessor to support further updates and provide a better overview of the preprocessor functionality.
- Updated the model in use from YOLOv8 to YOLOv11. 
- Optimized inference (object detection now takes 80 to 400 ms depending on the image size and complexity).
- Improved logging for inference speed tracking and debugging.
- Enabled automatic device choice based on GPU availability.
- Delegated image resizing and results filtering based on confidence threshold to YOLO for speed optimization.

Tested by running the Chrome Extension (Test) on 50+ images and comparing logs with the renderings page.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/preprocessors/yolov8/Dockerfile
+++ b/preprocessors/yolov8/Dockerfile
@@ -1,4 +1,4 @@
-FROM ultralytics/ultralytics:latest@sha256:af4572bf0df164d5708e4a1f50bddb5a74cb7d793fe46b7daf42b1768fbec608
+FROM ultralytics/ultralytics:latest@sha256:3d52e286bc345343b1a7a66dd38c44bb412bf418eb97d14298b8d8deb077f2e4
 
 WORKDIR /usr/src/app
 RUN adduser --disabled-password python 
@@ -15,7 +15,16 @@ RUN pip3 install --upgrade pip && \
 
 COPY /schemas /usr/src/app/schemas
 
-RUN wget https://image.a11y.mcgill.ca/models/objectDetectionWeights/yolov8x.pt
+# Set environment variable for model path
+ENV YOLO_MODEL_PATH=/usr/src/app/models/yolo11x.pt
+ENV CONF_THRESHOLD=0.75
+ENV MAX_IMAGE_SIZE=640
+
+# Create model directory
+RUN mkdir -p /usr/src/app/models
+
+# Download YOLOv11 model
+RUN wget -O /usr/src/app/models/yolo11x.pt https://unicorn.cim.mcgill.ca/image/resources/models/objectDetectionWeights/yolo11x.pt
 
 COPY /preprocessors/yolov8 /usr/src/app 
 
--- a/preprocessors/yolov8/Dockerfile
+++ b/preprocessors/yolov8/Dockerfile
@@ -24,7 +24,7 @@ ENV MAX_IMAGE_SIZE=640
 RUN mkdir -p /usr/src/app/models
 
 # Download YOLOv11 model
-RUN wget -O /usr/src/app/models/yolo11x.pt https://unicorn.cim.mcgill.ca/image/resources/models/objectDetectionWeights/yolo11x.pt
+RUN wget -O /usr/src/app/models/yolo11x.pt https://image.a11y.mcgill.ca/models/objectDetectionWeights/yolo11x.pt
 
 COPY /preprocessors/yolov8 /usr/src/app 
 
--- a/preprocessors/yolov8/README.md
+++ b/preprocessors/yolov8/README.md
@@ -1,20 +1,76 @@
+
+# Object Detection Preprocessor
+
 Beta quality: Useful enough for testing by end-users.
 
-This preprocessor detects the objects present in the image. The entire code for this preprocessor has been refered from [ultralytics Repository](https://github.com/ultralytics/ultralytics).
+This preprocessor detects objects present in images using YOLOv11. 
+The implementation builds upon the [Ultralytics](https://github.com/ultralytics/ultralytics) framework.
 
-In order to run this preprocessor we have used the [ultralytics YOLOv8 image](https://hub.docker.com/r/ultralytics/ultralytics) as a base docker image. The image is versioned with tag 'latest' and digest 'sha256:af4572bf0df164d5708e4a1f50bddb5a74cb7d793fe46b7daf42b1768fbec608'. The following commands import the base docker image:
+## Overview
 
-```docker build -t <image-name>``` 
+The preprocessor identifies various objects in images and returns their positions, classifications, and confidence scores in a standardized JSON format.
 
-```docker run -p <port-number>:5000 <image-name>```
+## Docker Setup
 
-The confidence threshold can be adjusted in the Dockerfile by changing the variable 'c_thres' at the top of detect.py
+This preprocessor uses the [Ultralytics image](https://hub.docker.com/r/ultralytics/ultralytics) as a base. The image is versioned with tag 'latest' and digest 'sha256:3d52e286bc345343b1a7a66dd38c44bb412bf418eb97d14298b8d8deb077f2e4'.
+
+To build and run the container:
+
+```bash
+docker build -t object-detection .
+docker run -p 5000:5000 object-detection
+```
+
+## Configuration
+
+The preprocessor can be configured using the following environment variables:
+
+| Variable | Description | Default Value |
+|----------|-------------|---------------|
+| YOLO_MODEL_PATH | Path to the YOLOv11 model file | /usr/src/app/models/yolo11x.pt |
+| CONF_THRESHOLD | Minimum confidence threshold for detection (0-1) | 0.75 |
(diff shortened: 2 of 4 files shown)


preprocessors/yolov8/detect.py
```diff
@@ -15,364 +15,207 @@
...
 app = Flask(__name__)
+
+# Configure logging
 logging.basicConfig(level=logging.DEBUG)
+log_pii = True
+
+# Environment variables and constants
+MODEL_PATH = os.environ.get(
+    'YOLO_MODEL_PATH',
+    'https://github.com/ultralytics/assets/releases/download/v8.3.0/yolo11x.pt'
```
jeffbl, line 42: If we're pulling from github, I think this should be logged at WARN level, since it isn't great to be doing lots of github pulls if someone is testing this and restarting it all the time.

preprocessors/yolov8/detect.py
```diff
@@ -15,364 +15,207 @@
...
+    )
+CONF_THRESHOLD = float(os.environ.get('CONF_THRESHOLD', '0.75'))
+MAX_IMAGE_SIZE = int(os.environ.get('MAX_IMAGE_SIZE', '640'))
+
+# Load the model
+model = YOLO(MODEL_PATH, verbose=log_pii)
```
jeffbl, line 48: please add comment explaining how this interacts with logging PII

preprocessors/yolov8/detect.py
```diff
@@ -15,364 +15,207 @@
...
+# Configure logging
 logging.basicConfig(level=logging.DEBUG)
+log_pii = True
+
+# Environment variables and constants
+MODEL_PATH = os.environ.get(
+    'YOLO_MODEL_PATH',
+    'https://github.com/ultralytics/assets/releases/download/v8.3.0/yolo11x.pt'
+    )
+CONF_THRESHOLD = float(os.environ.get('CONF_THRESHOLD', '0.75'))
```
jeffbl, line 44: These parameters should be logged, so we know from logs, for example, the confidence threshold for debugging.

preprocessors/yolov8/detect.py
```diff
@@ -15,364 +15,207 @@
...
+MAX_IMAGE_SIZE = int(os.environ.get('MAX_IMAGE_SIZE', '640'))
+
+# Load the model
+model = YOLO(MODEL_PATH, verbose=log_pii)
+
+# Choose GPU for processing if available
+device = 'cuda:0' if torch.cuda.is_available() else 'cpu'
```
jeffbl, line 51: Will this override the setting in docker-compose if we set it to another physical GPU there?

preprocessors/yolov8/detect.py
gvzdv: GitHub path is a fallback in case the system can't pull the model from `YOLO_MODEL_PATH` which points to the YOLO model stored on Pegasus.

preprocessors/yolov8/detect.py
gvzdv: At the moment, it is set to use the first available GPU.
Alternative options:
- assign it to another GPU. E.g., if we want to move it to the second GPU: `device = 'cuda:1' if torch.cuda.is_available() else 'cpu'`
- create logic that chooses a GPU that is less busy at the moment

Which option do you think works best in our case?

jeffbl: If the docker-compose setting only exposes the specified GPU to the container, then first available sounds correct to me. Just want to make sure that is how it works.

preprocessors/yolov8/detect.py
jeffbl: Understood. Not suggesting any different behavior, just that if it is pulling directly from github, that it show up in the logs to encourage using a local copy. We've had concerns in the past about being rate limited from pulling too many large files, etc. so we try and minimize where possible.

gvzdv: Do we need to have this fallback at all? Right now it fails because the container runs as a non-root user without write permissions.
Options:
- fix permissions, leave fallback URL, create a warning if the model is pulled from GitHub
- remove fallback, use Pegasus as the only place to pull the model from

preprocessors/yolov8/detect.py
gvzdv: At the moment, it doesn't specify the exposed GPU. I will add it using `device_ids`. Which one should it be?

jeffbl: The docker compose GPU specification only exposes the selected ones to the container, so this is fine.

jeffbl (approved):
Over to you @jaydeepsingh25

preprocessors/yolov8/Dockerfile
```diff
@@ -15,7 +15,16 @@ RUN pip3 install --upgrade pip && \
 
 COPY /schemas /usr/src/app/schemas
 
-RUN wget https://image.a11y.mcgill.ca/models/objectDetectionWeights/yolov8x.pt
+# Set environment variable for model path
+ENV YOLO_MODEL_PATH=/usr/src/app/models/yolo11x.pt
+ENV CONF_THRESHOLD=0.75
+ENV MAX_IMAGE_SIZE=640
```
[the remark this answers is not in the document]
gvzdv: You are absolutely correct. Fixed.

preprocessors/yolov8/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ultralytics/ultralytics:latest@sha256:af4572bf0df164d5708e4a1f50bddb5a74cb7d793fe46b7daf42b1768fbec608
+FROM ultralytics/ultralytics:latest@sha256:3d52e286bc345343b1a7a66dd38c44bb412bf418eb97d14298b8d8deb077f2e4
```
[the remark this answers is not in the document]
gvzdv: It was set up this way in the previous implementation, so I assumed there was a reason to go with the specific build.
Ultralytics [documentation](https://docs.ultralytics.com/guides/docker-quickstart/) suggests using `ultralytics/ultralytics:latest`.

preprocessors/yolov8/README.md
```diff
@@ -1,20 +1,76 @@
...
 
-In order to run this preprocessor we have used the [ultralytics YOLOv8 image](https://hub.docker.com/r/ultralytics/ultralytics) as a base docker image. The image is versioned with tag 'latest' and digest 'sha256:af4572bf0df164d5708e4a1f50bddb5a74cb7d793fe46b7daf42b1768fbec608'. The following commands import the base docker image:
+The preprocessor can be configured using the following environment variables:
 
-```docker build -t <image-name>``` 
+| Variable | Description | Default Value |
+|----------|-------------|---------------|
+| YOLO_MODEL_PATH | Path to the YOLOv11 model file | /usr/src/app/models/yolo11x.pt |
+| CONF_THRESHOLD | Minimum confidence threshold for detection (0-1) | 0.75 |
+| MAX_IMAGE_SIZE | Maximum image dimension for processing | 640 |
```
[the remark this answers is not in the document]
gvzdv: Yes, this is the size in pixels. It's the highest resolution supported by YOLO at the moment.
YOLO will automatically resize images using the larger side of the image and maintaining the aspect ratio.

preprocessors/yolov8/detect.py
```diff
@@ -15,364 +15,214 @@
...
+else:
+    device, device_name = 'cpu', 'CPU'
+
+# Load schemas once at startup
+with open('./schemas/preprocessors/object-detection.schema.json') as f:
+    DATA_SCHEMA = json.load(f)
+with open('./schemas/preprocessor-response.schema.json') as f:
+    RESPONSE_SCHEMA = json.load(f)
+with open('./schemas/definitions.json') as f:
+    DEFINITIONS_SCHEMA = json.load(f)
```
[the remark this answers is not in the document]
jeffbl: @jaydeepsingh25 this is left as-is for now, but please ping @gvzdv to discuss if you think we need to set a different standard here. (I'm not weighing in one way or the other)

preprocessors/yolov8/detect.py
jeffbl: (github fallback is removed)

preprocessors/yolov8/detect.py
jeffbl: (logging added)

PR #995, opened 2025-03-30
fix: resize image util

### LINKED ISSUE #867: Create standard way for preprocessors to do graphic resizing
Talking with @AndyBaiMQC this morning, it will likely be necessary to resize graphics going to local ML models, each of which may have its own constraints. We've been burned by resizing issues numerous times throughout the project ([example](https://github.com/Shared-Reality-Lab/IMAGE-server/issues/610) [example](https://github.com/Shared-Reality-Lab/IMAGE-server/issues/348) example). Goal of this work item is to come up with a clear and documented way for preprocessors to resize graphics. Key functionality includes:

- Must support a wide variety of graphic formats (jpg, png, gif, etc.)
- good performance
- easy to integrate (easy, clean API that is hard to mess up)

Assigning to @JRegimbal first to weigh in on architecture, before going to @jaydeepsingh25 for implementation. Some options:

- Choose a library and have each preprocessor use it directly
- Implement a service that encapsulates resizing so that if we want to change it in the future, we only have to change it in one place (e.g., to support a new graphic format, or if the library we initially choose turns out to have flaws or is no longer supported, etc)
- Have the orchestrator resize before sending to preprocessor based on a parameter in docker-compose
- ???

BONUS: If we centralize this, we could also use it to transform photos into a common format, to support a wider variety of weird graphic types. E.g., if the library can transform .png into .jpg as well, then we don't have to rely on preprocessors supporting .png. Probably should break out into separate issue, or generalize this one into "transform graphics before they go to preprocessors)

jaydeepsingh25:
Resolves #867 . This pull request includes changes to add a new image resizing function. The most important changes are summarized below:

### New Image Resizing Function:

* [`config/image-resize.py`](diffhunk://#diff-8f2de57a5144d5cd34a3f447e6dd49b746f7cab06917387ab3909eacb3fcb056R1-R54): Added a new function `resize_image` that resizes a base64 encoded image to a specified size, with an option to maintain the aspect ratio. The function includes logging for the original and new image sizes and formats, and handles conversion of GIF images to PNG.

#### Key changes:

* Added a new function `resize_image` to resize base64 encoded images, with an option to maintain the aspect ratio.
* Included logging to track the original and new image sizes and formats, as well as to handle any errors that occur during the resizing process.
* Implemented conversion of GIF images to PNG format before resizing.

Tested on different image sizes and formats: 

- Resize png image without aspect ratio: 
image

- Resize jpeg maintaining aspect ratio:
image

- Resize gif without aspect ratio:
image

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

OR
This pull request introduces a new function to resize base64 encoded images in the `config/image-resize.py` file. The function includes logging for debugging and can handle GIF images by converting them to PNG format.


config/image-resize.py
```diff
@@ -0,0 +1,54 @@
...
+from io import BytesIO
+import logging
+
+
+def resize_image(base64_image_str, size, maintain_aspect_ratio=False):
+    """
+    Resize a base64 encoded image to the specified size.
+
+    :param base64_image_str: Base64 encoded image string
+    :param size: Tuple (width, height) specifying the new size
```
jeffbl, line 12: Suggest adding maintain_aspect_ratio here, and setting to True by default, since I suspect this will be more commonly desired.

config/image-resize.py
```diff
@@ -0,0 +1,54 @@
...
+        format = image.format
+        logging.info(f'Original size - width: {width}, height: {height}')
+        logging.info(f'Original format: {format}')
+
+        # Check if the image is a GIF and convert to PNG
+        if image.format == "GIF":
+            logging.info("Image is a GIF. Converting to PNG.")
+            image = image.convert("RGBA")
+
+        if maintain_aspect_ratio:
```
jeffbl, line 30: Will this resize to be *larger* if the size parameter is bigger than the input graphic? For our purposes, this would almost always be undesirable, since it wouldn't add any additional content for ML to chew on, but potentially consume more space/time. Suggest either making this only resize smaller, or if we think someone might have a valid use case for making it larger, make it an additional parameter that can be set.

jeffbl (changes requested):
Glad we'll have a common way to do this! In addition to a few specific comments/questions:

1. Is PNG the best common format, vs. something like jpeg? Want to confirm we're choosing something that will be usable by the largest possible set of downstream tools.
2. If I only want to map a graphic to another format, without resizing, how would I do that? (i.e., we know GIF graphics are problematic, so if we decided to convert *all* graphics to PNG or another common format, but leave resizing up to the individual components, can I do that?)

gvzdv:
> Glad we'll have a common way to do this! In addition to a few specific comments/questions:
> 
> 1. Is PNG the best common format, vs. something like jpeg? Want to confirm we're choosing something that will be usable by the largest possible set of downstream tools.
> 2. If I only want to map a graphic to another format, without resizing, how would I do that? (i.e., we know GIF graphics are problematic, so if we decided to convert _all_ graphics to PNG or another common format, but leave resizing up to the individual components, can I do that?)

1. I would vote for PNG as it uses lossless compression (as opposed to lossy JPEG), and we don't lose (potentially valuable) data during the file transformation. Need to check if all downstream preprocessors are okay with that (I am 99% sure they are, as we're currently testing the system mostly on JPEG and PNG images).
2. Maybe it makes sense to split this into two different utilities? Unify the format for all (or all GIF) images and resize them only if needed.

I believe that transforming images (either only from "problematic" formats like GIF or all of them) is a better approach than looking for a solution for each individual preprocessor, given that the preprocessors won't be able to work with anything beyond the first frame of the GIF image anyway.

config/process_image.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # Determine output format
+        target_format = output_format if output_format else original_format
+
+        # Process the image
+        if needs_resize:
+            logging.info("Image needs resizing (maintaining aspect ratio).")
+            image.thumbnail(output_size)
+        else:
+            logging.info("Image dimensions are within limits, no resizing "
```
jeffbl, line 49: If it did not need resize, and input and output formats are the same, can we just return the original here, rather than recompressing it below, which could potentially reduce quality due to the recompress?

config/process_image.py
```diff
@@ -0,0 +1,72 @@
...
+    The image is only resized if any dimension exceeds the specified
+    output_size.
+    Aspect ratio is always maintained during resizing.
+
+    :param base64_image_str: Base64 encoded image string
+    :param output_size: Maximum size (width, height) for the output image
+    :param output_format: Optional format to convert the image to
+                         (e.g., 'PNG', 'JPEG')
+                         If None, the original format is preserved
+    :return: Processed image or original image object in case of an error
```
jeffbl, line 21: Suggest noting that it returns an image, but not base64

jeffbl (approved):
Looks ok, with a couple caveats. I'm fine with merging it as-is, with issues handled later, so that @JRegimbal can implement request-preprocessor example asap.

config/process_image.py
```diff
@@ -0,0 +1,72 @@
...
+            f'Final size - width: {final_image.width}, '
+            f'height: {final_image.height}'
+        )
+        logging.info(f'Final image format: {final_image.format}')
+
+        return final_image
+
+    except Exception:
+        logging.exception("Image processing error")
+        # Return original image if it exists
```
JRegimbal, line 71: Would recommend catching more specific errors or logging a useful message. I'm not sure how helpful this would be for debugging.

jeffbl:
@JRegimbal if you're ok merging as-is so that we can get the initial implementation done, please go ahead, and we can make the above changes after. Don't want you to be blocked if you are still trying to complete the request rewrite before you're offline.

Issue #1010, opened 2025-04-10
orchestrator logs "500 INTERNAL SERVER ERROR" and returns empty list of interpretations


jeffbl:
Chrome on Linux, test extension, against unicorn.
Graphic: https://onlinejpgtools.com/images/examples-onlinejpgtools/product-manager-business-card.jpg

First time tried it, it returned an empty list of experiences, which should not happen:

![Image](https://github.com/user-attachments/assets/adaa1199-7eba-492b-9aaf-e4d08b8c7a55)

Logs show orchestrator giving "500 INTERNAL SERVER ERROR"

```
orchestrator-1                    | 2025-04-10T15:43:04.103602017Z Received request                                                                                                                                                                           
orchestrator-1                    | 2025-04-10T15:43:04.104218190Z No route defined in request. Setting default value.                                                                                                                                        
orchestrator-1                    | 2025-04-10T15:43:04.121169387Z Using default route ("default") for object-detection                                                                                                                                       
orchestrator-1                    | 2025-04-10T15:43:04.121193092Z Using default route ("default") for object-grouping                                                                                                                                        
orchestrator-1                    | 2025-04-10T15:43:04.121197620Z Using default route ("default") for object-sorting                                                                                                                                         
orchestrator-1                    | 2025-04-10T15:43:04.121201778Z Using default route ("default") for graphic-caption                                                                                                                                        
orchestrator-1                    | 2025-04-10T15:43:04.121207739Z Using default route ("default") for openstreetmap                                                                                                                                          
orchestrator-1                    | 2025-04-10T15:43:04.121211586Z Using default route ("default") for semantic-segmentation                                                                                                                                  
orchestrator-1                    | 2025-04-10T15:43:04.121223028Z Using default route ("default") for line-charts                                                                                                                                            
orchestrator-1                    | 2025-04-10T15:43:04.121234359Z Using default route ("default") for action-recognition                                                                                                                                     
orchestrator-1                    | 2025-04-10T15:43:04.121328375Z text-followup has routes followup                                                                                                                                                          
orchestrator-1                    | 2025-04-10T15:43:04.121343433Z Using default route ("default") for content-categoriser                                                                                                                                    
orchestrator-1                    | 2025-04-10T15:43:04.121347441Z Using default route ("default") for autour-preprocessor                                                                                                                                    
[...]

PR #1014, opened 2025-04-15
Improve handler error reporting in the orchestrator

JRegimbal:
As mentioned in #1010, the current way that the orchestrator reports HTTP error codes returned by handlers suggests that the orchestrator itself is experiencing the error. This updates the log line to clearly indicate which handler the error code originates in. For example:
```
orchestrator-1  | Waiting for handlers...
orchestrator-1  | Received 500 INTERNAL SERVER ERROR from svg-od-handler.
orchestrator-1  | Failed to validate the response renderer
orchestrator-1  | Valid response generated.
```
Previously the line would read "500 INTERNAL SERVER ERROR".

Tested with sample graphics on unicorn and there does not appear to be any regression from current main. Assigning to @jeffbl for feedback on the error message before merging.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/orchestrator/src/server.ts
+++ b/orchestrator/src/server.ts
@@ -396,7 +396,7 @@ app.post("/render", (req: express.Request, res: express.Response) => {
                     if (resp.ok) {
                         return resp.json() as Promise<HandlerResponse>;
                     } else {
-                        console.error(`${resp.status} ${resp.statusText}`);
+                        console.error(`Received ${resp.status} ${resp.statusText} from ${handler[0]}.`);
                         const result = await resp.json();
                         throw result;
                     }


jeffbl (approved):
This makes the message clearer, but we still have two problems:
1. why was a preprocessor having an internal server error?
2. why was the orchestrator allowing a blank list of experiences, rather than just returning the "IMAGE can't process this" message?

JRegimbal:
1) A preprocessor was not. One of the debug handlers seems to not behave well when the data it expects is not present.
2) Not sure. Considering I couldn't replicate, I would guess that something is going wrong in the extension.

Anything on the messages or can I merge?

PR #1042, opened 2025-05-05
Fix hex-coded characters in Ollama responses

gvzdv:
Characters like `&#x27` were found in responses from Ollama.
This PR fixes the issue:
- `html.unescape()` converts characters from the HTML code
- `jsonify` ensures special characters are escaped while constructing the response object

Tested by forcing Ollama to generate text with quotation marks and apostrophes and checking the text representation in the extension and logs.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/preprocessors/graphic-caption/caption.py
+++ b/preprocessors/graphic-caption/caption.py
@@ -21,6 +21,7 @@ import time
 import jsonschema
 import logging
 import os
+import html
 from datetime import datetime
 from config.logging_utils import configure_logging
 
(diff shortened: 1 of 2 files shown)


jeffbl:
We're not merging this before deploying to production, since it does not actually fix the issue (core issue appears to be in the TAT). Still a good thing to do, but let's test on unicorn first before rolling out to production, since this impacts more than just the Monarch-specific functionality.

jeffbl:
No red flags on my side. We can merge for testing now that demo is over. Assigning to @jaydeepsingh25 as the proper reviewer for preprocessors now that we're out of demo crunch mode.

gvzdv:
It turned out that the HTML characters were generated on the handler/TAT side, so this PR is non-essential since LLMs very rarely produce HTML characters.

jeffbl:
Thanks! Do you still recommend merging, since it is still an outside possibility? Or is there a downside, and this should be closed without merging?

gvzdv:
I don't see major downsides, and it covers the 0.001% possibility of the model generating the HTML characters, so I suggest merging.

PR #1065, opened 2025-05-30
Let pseudo-preprocessors modify requests

### LINKED ISSUE #1008: Create way to modify request before it goes to preprocessors/handlers
When a request comes in, we are finding a need to modify certain parts of it to change formats. The two examples so far are:

1. if we get a google placeid, this needs to be turned into a lat/lng before being passed along. (@JRegimbal please modify if this is incorrect)
2. we want to standardize the internal format for photos to something like PNG, so that all preprocessors/handlers know they will receive this format.

Currently, we don't have a disciplined and extensible way to do this. Goal of work item is to:

- Come up with a good way of doing this, such that it can be extended in the future
- Make sure that the resizing/reformatting code library is taken into account (#867) Can we use the same library somehow? Does the API support everything we need?

A "purer" way of doing this could also be to simply create a new preprocessor that is first in the run group (maybe even hardcoded), that returns a modified request, giving everything downstream access to either the original content unmodified, or the "revised" request that standardizes formats, etc. But leaving it up to @JRegimbal to decide the cleanest way to accomplish the goals.

### LINKED ISSUE #1028: Allow certain preprocessors to directly modify request data
## Context

Situations exist where adding new data to a request, as preprocessors do currently, is not appropriate. Consider:

1. A web graphic is sent as a GIF and certain preprocessors are unable to work with it. #960 This is an edge case preprocessor designers should not need to consider.
2. A location has a place ID but no geospatial coordinates. A handler assumes that such coordinates will always be present and will either error, or need to reconstruct the information itself.

In these cases, adding data in the `preprocessors' object is inappropriate as this allows for redundancy with fields already allowed in the request. Rather than require several levels of checking for such basic information within each preprocessor and handler, it would be smoother if preprocessors were able to modify the original request to make the information therein more usable.

## Proposal

Preprocessors that modify a request should never run after typical ones that add new data, since this may create race conditions. Consider how terrible debugging would be if the graphic that is the subject of a request changed partway through the service graph!

I suggest *adding a new label* indicating to the orchestrator that a preprocessor will modify the request. For example, `ca.mcgill.a11y.image.preprocessor.modifyRequest: true`. These preprocessors will support all other labels preprocessors normally have.

Upon receiving a request, the orchestrator will construct two separate graphs: one where `modifyRequest` is true, and the rest. It will run the `modifyRequest` graph first, allowing for dependencies to exist between these special preprocessors. Dependencies could not exist between these preprocessors and regular ones that run later. Since the graphs are constructed separately, a `modifyRequest` preprocessor that depends on a normal one would be treated as if it depended on a preprocessor that is not running, and would be excluded from the graph.

These preprocessors would also use the same endpoint and response structure as all other preprocessors. Rather than returning a specific preprocessor data type, it would respond with a the keys it wishes to add or modify within the request. For example, if a preprocessor was meant to convert a graphic into a PNG, it would return the following to the orchestrator:
```json
{
  "request_uuid": "$REQUEST_UUID",
  "timestamp": 123456789,
  "name": "ca.mcgill.a11y.image.request",
  "data": {
    "graphic": "data:image/png;..."
  }
}
```
`modifyRequest` preprocessors may not modify the timestamp or request_uuid of the request itself. If a `modifyRequest` preprocessor replies with a `name` field that is not `ca.mcgill.a11y.image.request`, the orchestrator should reject it as invalid. Similarly, if a preprocessor without the `modifyRequest' label uses this reverse domain ID, the orchestrator should reject its response.

Note that while `modifyRequest` preprocessors can depend on one another, they are not aware of which request fields the others are modifying. Care must still be taken here, as with other preprocessors, to ensure that write-after-write hazards do not occur. See #1020 for a related discussion.

Pinging @jeffbl @shahdyousefak @jaydeepsingh25 for comment.

JRegimbal:
No, I'm not married to the name "pseudo-preprocessors".

Resolve #1008, resolve #1028. Allows for preprocessors to modify the request if:
* They have the label `ca.mcgill.a11y.image.modifyRequest: true`
* They use the name `ca.mcgill.a11y.image.request`.

Certain fields in the request, such as `request_uuid`, cannot be modified.

Tested using the graph-based scheduler and the priority based one. It seems to be working with a toy pseudo-preprocessor.

Note that caching currently does not and **cannot** work with pseudo-preprocessors as implemented. Data is retrieved from the cache using the name in the response, which for these will always be `ca.mcgill.a11y.image.request`. I have no idea how you would reasonably handle rewrites to the same field or fields here, and am open to discussion. Assigning @jeffbl and @shahdyousefak to look at the code, and including @jaydeepsingh25 for thoughts on caching. Feel free to un-assign yourselves once you've left a comment.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/orchestrator/src/docker.ts
+++ b/orchestrator/src/docker.ts
@@ -14,6 +14,7 @@
  * and our Additional Terms along with this program.
  * If not, see <https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/LICENSE>.
  */
+import { PreprocessorInfo } from "./server";
 import Docker from "dockerode";
 
 export const docker = new Docker();
@@ -27,6 +28,7 @@ const _CACHE_TIMEOUT_LABEL = "ca.mcgill.a11y.image.cacheTimeout";
 const _ORCHESTRATOR_ID_ = process.env.HOSTNAME as string;
 export const DEFAULT_ROUTE_NAME = "default";
 const _ROUTE_LABEL_ = "ca.mcgill.a11y.image.route";
+const _MODIFY_REQUEST_LABEL_ = "ca.mcgill.a11y.image.modifyRequest";
 
 function getOrchestratorNetworks(containers: Docker.ContainerInfo[]): string[] {
     const container = containers.filter(c => c.Id.startsWith(_ORCHESTRATOR_ID_)).shift();
@@ -86,6 +88,7 @@ export function getPreprocessorServices(containers: Docker.ContainerInfo[], rout
         const portLabel = container.Labels[_PORT_LABEL_];
         const priorityLabel = Number(container.Labels[_PREPROCESSOR_LABEL_]);
         const cacheTimeout = Number(container.Labels[_CACHE_TIMEOUT_LABEL] || 0);
+        const modifyRequest = Boolean(container.Labels[_MODIFY_REQUEST_LABEL_] || false);
 
         let port;
         if (portLabel !== undefined) {
@@ -96,7 +99,7 @@ export function getPreprocessorServices(containers: Docker.ContainerInfo[], rout
         } else {
             port = 80;
         }
-        return [container.Labels["com.docker.compose.service"], port, priorityLabel, cacheTimeout];
+        return [container.Labels["com.docker.compose.service"], port, priorityLabel, cacheTimeout, modifyRequest];
     });
 }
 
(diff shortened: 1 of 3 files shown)


jeffbl:
No comments other than name needs to change. tyvm for implementing. Need to discuss with @jaydeepsingh25 if he is blocked on #995 in any way, since that will be necessary before we can implement the graphic format/resize pseudopreprocessor.

PR #1103, opened 2025-07-03
Override the cache

### LINKED ISSUE #1102: Create a cache override feature in the orchestrator
There are cases where we want to specify what will be returned by a preprocessor to ensure consistency, e.g., in the context of a user study. In the past, we have either presented renderings outside the context of an IMAGE client, or overridden preprocessors to return particular responses when given certain inputs. Neither solution is ideal, requiring considerable work from the team.

This issue proposes to modify how the orchestrator retrieves cached information. Before checking the actual cache for a value corresponding to the hashed request, the orchestrator would look for the hash in a file or files of manually set responses. If the hash is present, the manual response is returned without the cache being consulted. If no hash is present, the cache will be checked as usual. This method should require minimal changes to orchestration logic by modifying caching functions in the orchestrator

JRegimbal:
Resolves #1102. This is a work in progress. While the code currently works, it would not be easy to use. Any solution that relies on knowing the cache hash is not workable since the resize-graphic pseudo-preprocessor will modify `graphic`. This value is used to calculate the hash, meaning different hashes will be obtained if run locally.

Sharing this for comment, but trying to think of a better solution.

EDIT: While terrible, it is sufficient for our needs now. I will think of something better later. Tested with environment variable unset, with variable set but no override, and with an override. All appears to work as anticipated.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/orchestrator/src/server.ts
+++ b/orchestrator/src/server.ts
@@ -106,6 +106,21 @@ async function measureExecutionTime<T>(label: string, fn: () => Promise<T>): Pro
 }
 
 async function checkCache(preprocessorName: string, hashedKey: string, cacheTimeOut: number): Promise<Response | null> {
+    if (process.env.CACHE_OVERRIDE != undefined && preprocessorName) {
+       const filepath = path.join(process.env.CACHE_OVERRIDE, hashedKey);
+       try {
+            // Load cache override and serve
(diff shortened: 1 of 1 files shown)


JRegimbal:
@VenissaCarolQuadros @jeffbl thoughts on alternatives based on the above?

PR #1115, opened 2025-07-14
imageup to run on Pegasus (production), just not when triggered by cron 

shahdyousefak:
We want imageup to run on production (Pegasus), but we do not want it to be run automatically via image_daily_cron on production. Now, we are relying on the cron script itself (image_daily_cron) to control where it runs (test machines only).

imageup --auto will not run on production servers, but manual runs of imageup (without --auto) will still work on production just fine.

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/scripts/imageup
+++ b/scripts/imageup
@@ -27,6 +27,12 @@ if [[ -z "$COMPOSE_ENV_FILE" || -z "$ENVIRONMENT" ]]; then
   exit 1
 fi
 
+# Skip automatic daily resets in production (we use Pegasus)
+if [[ "$1" == "--auto" && "$ENVIRONMENT" == "production" ]]; then
+  echo "$(date) - Skipping imageup: auto-run skipped in production" >> /tmp/imageup_skip.log
+  exit 0
+fi
+
 delete_override=true  # default to deleting override file
 
 # get command line arguments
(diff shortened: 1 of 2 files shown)


scripts/imageup
```diff
@@ -27,6 +27,12 @@ if [[ -z "$COMPOSE_ENV_FILE" || -z "$ENVIRONMENT" ]]; then
   exit 1
 fi
 
+# Skip automatic daily resets in production (we use Pegasus)
+if [[ "$1" == "--auto" && "$ENVIRONMENT" == "production" ]]; then
+  echo "$(date) - Skipping imageup: auto-run skipped in production" >> /tmp/imageup_skip.log
```
jeffbl, line 32: why is there a separate logfile for just this, rather than a general log that capture the full results?

scripts/imageup
```diff
@@ -27,6 +27,12 @@ if [[ -z "$COMPOSE_ENV_FILE" || -z "$ENVIRONMENT" ]]; then
   exit 1
 fi
 
+# Skip automatic daily resets in production (we use Pegasus)
+if [[ "$1" == "--auto" && "$ENVIRONMENT" == "production" ]]; then
```
jeffbl, line 31: suggest using an actual flag tool like we do in imageup elsewhere, since if someone does `imageup -k --auto, this will fail...`

jeffbl (approved):
Only issue with new argument parsing is that multiple single character arguments cannot be combined. But this is an internal script, so not going to insist on this.

PR #1120, opened 2025-07-23
Switch to using the service name for cache

### LINKED ISSUE #855: Caching doesn't serve from cache until third request
restoreunstable to clear cache

Go to a photo on IMAGE homepage
Get IMAGE rendering
Note in logs it is storing it in cache

Get rendering for same photo
NOTE: stores results in cache
EXPECT: serves results from cache since it just stored them

Get rendering for same photo
NOTE: correctly serves from cache

Get IMAGE rendering for a photo2 (different photo)
NOTE: stores in cache

Get IMAGE rendering for photo2
NOTE: correctly serves from cache

Interpretation: once one photo is cached successfully twice, future photos cache correctly on first try

Go to map, e.g. https://developers.google.com/maps/documentation/embed/get-started
Interpret this map with IMAGE
NOTE: stores autour results in cache

Interpret this map with IMAGE
NOTE: stores again in cache
EXPECT: serves from cache since it should already be stored

Interpret this map with IMAGE
NOTE: serves from cache (correct)

So other request types or something with which preprocessors are getting run (or some other issue) means that the first time things are cached, they may or may not "stick" the first time, but work after being cached a second time.

EXPECT: Caching is successful the first time, and second request reliably returns results from cache.

### LINKED ISSUE #1104: Cache override not working on first request
Cache override is only fetching from the cached responses from the second run. The first run yields Qwen responses.

1. Run imageup
2. Select 'Load in Tactile Authoring Tool' for the last graphic in https://image.a11y.mcgill.ca/pages/multistage_diagrams.html (Jellyfish diagram). This yields results as shown below. (Qwen)
Image
3. Repeating step 2 gives cached Gemini responses.
Image

Note: it does not need to be the same graphic between step 2 and 3 (i.e. step 2 fetches Qwen results and step 3 fetches cache override results even if I try the human evolution in step 2 and jellyfish life cycle in step 3). 
As @jeffbl pointed out, this is very similar to cache not working for the first graphic of a certain type. 

I also tried making a request on a [cat photograph](https://i.guim.co.uk/img/media/327aa3f0c3b8e40ab03b4ae80319064e401c6fbc/377_133_3542_2834/master/3542.jpg?width=1200&height=1200&quality=85&auto=format&fit=crop&s=34d32522f47e4a67286f9894fc81c863) between step 1 and 2 and confirmed that the image is fetched from cache on the third request  before proceeding to step 2 but I still see the behavior described above.

### LINKED ISSUE #1116: Switch cache to use service name rather than preprocessor response ID
In the same way graph-based parallel execution in the orchestrator uses the service names to identify dependencies, the service name should be used to calculate the hash used for caching. This should resolve #855 and #1104.

JRegimbal:
Rather than use the name of the preprocessor data type, the cache now indexes using the name of the service. This resolves #855, resolves #1104, and resolves #1116. Tested by clearing the cache and sending photo requests, and by creating a manual override. Note that the cached data now contains the data type and preprocessor data (e.g., `{"name": "ca.mcgill.a11y.image.preprocessor.objectDetection", "data": {...}}`).

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/orchestrator/src/server-cache.ts
+++ b/orchestrator/src/server-cache.ts
@@ -41,10 +41,10 @@ export class ServerCache {
 
     }
     
-    async setResponseInCache(hashedKey: string, value: string, timeout: number){
-        console.debug(`storing data in memcache with key ${hashedKey}`);
+    async setResponseInCache(hashedKey: string, preprocessorType: string, value: Record<string, unknown>, timeout: number){
+        console.debug(`storing data in memcache with key ${hashedKey}, name ${preprocessorType}`);
         try{
-            await this.memjsClient.set(hashedKey, value, {expires: timeout});
+            await this.memjsClient.set(hashedKey, JSON.stringify({name: preprocessorType, data: value}), {expires: timeout});
         } catch(error){
             console.debug("Error setting response in the cache");
         }
--- a/orchestrator/src/server.ts
+++ b/orchestrator/src/server.ts
@@ -51,8 +51,6 @@ const serverCache = new ServerCache();
 const memjsClient = serverCache.memjsClient;
 
 console.debug("memcached server", memjsClient.servers);
-// variable to store mapping of service (as defined in docker-compose) and preprocessor-id (as returned in the reponse)
-const SERVICE_PREPROCESSOR_MAP : Record<string, string> = {};
 const port = 8080;
 const ajv = new Ajv2020({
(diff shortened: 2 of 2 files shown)


jeffbl (approved):
No red flags. Looking forward to testing, since this has been an ongoing weirdness!

jaydeepsingh25 (approved):

PR #1125, opened 2025-08-06
Return 204 if no stages found in multistage-diagram-segmentation

### LINKED ISSUE #1124: Multistage preprocessor returning empty arrays
When multistage preprocessor is enabled on graphics it cannot generate valid responses for, it returns empty arrays.

Repro steps:
1. Get IMAGE renderings for bear image on https://unicorn.cim.mcgill.ca/image/pages/tutorial.html (with Monarch enabled)
2. The response from multistage diagram preprocessor is as follows
`{'links': [], 'stages': []}` (This causes downstream elements dependent on this preprocessor response to run) 

While this preprocessor should not be running for this graphic (and could be fixed by the origin filtering), in the future, we might encounter cases where the preprocessor returns 0 length arrays even when responses are expected. In such cases, it would be ideal to return 204 instead of an empty array to ensure that the downstream elements are not triggered. [refer: preprocessor-chaining-best-practices on our wiki](https://github.com/Shared-Reality-Lab/IMAGE-server/wiki/2.-Handlers,-Preprocessors-and-Services#preprocessor-chaining-best-practices)

@gvzdv please let me know if you would recommend making schema changes as well to enforce the minimum length of 1 for stages array (or both stages and links arrays)

gvzdv:
Fixes #1124 

If the `multistage-diagram-segmentation` does not find any stages, the preprocessor will return the code `204` instead of returning an empty list of stages.

Tested on Unicorn using graphics that do not contain stages.
---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
+++ b/preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
@@ -702,7 +702,7 @@ def process_diagram():
             if isinstance(stage, dict) and "label" in stage
             ]
         if not stages or len(stages) == 0:
-            logging.warning(
+            logging.info(
                 "No stage labels found. Cannot request bounding boxes."
                 )
             return jsonify(
--- a/preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
+++ b/preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
@@ -701,14 +701,13 @@ def process_diagram():
             for stage in base_json.get("stages", [])
             if isinstance(stage, dict) and "label" in stage
             ]
-        if not stages:
+        if not stages or len(stages) == 0:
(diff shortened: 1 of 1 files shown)


preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
```diff
@@ -701,14 +701,13 @@ def process_diagram():
             for stage in base_json.get("stages", [])
             if isinstance(stage, dict) and "label" in stage
             ]
-        if not stages:
+        if not stages or len(stages) == 0:
             logging.warning(
```
jeffbl, line 705: not really warning level

jeffbl (changes requested):
log level warning doesn't seem appropriate, but otherwise looks good.

jeffbl (approved):

PR #1132, opened 2025-08-12
check GPU health at the *host* level, removing this outside of semseg

### LINKED ISSUE #1112: semseg preprocessor healthcheck failing, even though semseg works
semseg appears to work when I tried it a while back, but the health check keeps coming up as failing. This means we basically end up ignoring healthcheck notifications, since they happen all the time. Seem to happen starting about 11-19h, after reset happens at 7am. So fine for a while, then goes bad until next reset...

Image

shahdyousefak:
Currently, the semseg /health/gpu endpoint ran host checks from inside the container:
nvidia-smi (NVML)
/proc/driver/nvidia/version 

which was designed to compare the strings and return 500 on a mismatch or NVML error.
In practice, we found that containers can lose access to NVML/CUDA (or have it partially mounted) even when the host is fine. 
inside the container nvidia-smi -> “Failed to initialize NVML” and torch.cuda.is_available() -> False.

The scope was incorrect, since this is a host concern. Doing it inside the container makes results dependent on other variables besides the real driver state. 

I've updated the healthcheck script to include identical GPU checks to the host. if there’s a problem, it logs the precise reason and appends one host-level line to Slack.

This should fix the spam/false-positive pain; we parse numeric versions and just log them, we see exactly which versions caused the mismatch. 

Don't delete below this line.

---

## Required Information

- [X] I referenced the issue addressed in this PR.

## Coding/Commit Requirements

OR

### DIFF
--- a/preprocessors/mmsemseg/Dockerfile
+++ b/preprocessors/mmsemseg/Dockerfile
@@ -53,6 +53,6 @@ ENV FLASK_APP=segment.py
 USER python
 
 HEALTHCHECK --interval=60s --timeout=10s --start-period=120s --retries=5 \
-  CMD curl -f http://localhost:5000/health && curl -f http://localhost:5000/health/gpu || exit 1
+  CMD curl -f http://localhost:5000/health || exit 1
 
 CMD [ "gunicorn", "segment:app", "-b", "0.0.0.0:5000", "--capture-output", "--log-level=debug" ]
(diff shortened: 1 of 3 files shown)


JRegimbal (commented):
It looks good! Have you tested this on unicorn yet?

JRegimbal (approved):

PR #1136, opened 2025-08-19
Update content categorisation process

gvzdv:
Resolves #1015.

Updated the `content-categoriser` and the corresponding schema to produce boolean tags for the following categories:
```
[
        "photo",
        "diagram",
        "flow_diagram",
        "contains_text",
        "people",
        "animals",
        "collage",
        "chart_or_graph",
        "illustration"
    ]
```
The new schema is passed to the LLM alongside the updated prompt that includes the instructions and categories (this is redundant and requires prompt updates on schema updates, but improves the accuracy).

Examples:

1. 
image

Graphic category JSON: {'photo': True, 'diagram': False, 'flow_diagram': False, 'contains_text': False, 'people': False, 'animals': True, 'collage': False, 'chart_or_graph': False, 'illustration': False}

---

2.
image

Graphic category JSON: {'photo': False, 'diagram': True, 'flow_diagram': True, 'contains_text': True, 'people': False, 'animals': True, 'collage': False, 'chart_or_graph': False, 'illustration': True}

---

3.
image

Graphic category JSON: {'photo': False, 'diagram': True, 'flow_diagram': False, 'contains_text': True, 'people': False, 'animals': False, 'collage': False, 'chart_or_graph': True, 'illustration': False}

---

4.
image

Graphic category JSON: {'photo': False, 'diagram': False, 'flow_diagram': False, 'contains_text': True, 'people': False, 'animals': False, 'collage': True, 'chart_or_graph': False, 'illustration': True}

---

Note: this PR will require some cleanup after we discuss the following questions.
1. Do we need to add/remove any categories?
2. Do we need to structure the schema/output in a different way (e.g. construct it as it was constructed previously: `{"category": graphic_category_output}`)?
3. The model is generally good with the categorization (i.e. it identified the collage in example 4 when our custom collage detector did not), but it usually tags `diagram` and `chart_or_graph` in the same way (both `True` or both `False`). I agree that sometimes it's hard to distinguish between these categories. Should we keep them separate and update the description to improve the accuracy, or combine them in one category (`data_visualization_or_diagram`)?

---

## Required Information

- [x] I referenced the issue addressed in this PR.
- [x] I described the changes made and how these address the issue.
- [x] I described how I tested these changes.

## Coding/Commit Requirements

* [x] I followed applicable coding standards where appropriate (e.g., [PEP8](https://pep8.org/))
* [x] I have not committed any models or other large files.

## New Component Checklist (**mandatory** for new microservices)

[...]

PR #1156, opened 2025-09-25
fix: content categoriser updates

### LINKED ISSUE #1015: EPIC: As an IMAGE developer, I want a clean and consistent way to access a larger set of graphic tags, so that I can more easily make sense of incoming graphics.
Right now, we have separate preprocessors for deciding things like whether a photo is a collage, what type of graphic is is (chart, text, etc), and other aspects, all of which are essentially tags on the graphic. Although this means each of the tags can be treated independently in terms of implementation, this approach has multiple downsides:

1. Since we're using an LLM to do much of this tagging, the LLM is called multiple times independently, which is inefficient and slows down requests vs. just asking the LLM once for a group of tags.
2. Each tag has its own schema, which is complex and has significant overhead, plus each preprocessor needs to know the exact tag it is seeking and the relevant schema.

PROPOSAL: change to a more generic tagging approach, where new tags can be added more easily (i.e., without having to generate a whole new schema, and without having to implement an entirely new preprocessor). Implement in a way that multiple questions can be posed to an LLM in a single batch.

NOTES:
- if any tags depend on other tags, need a way to resolve this.
- need to decide on how tags are treated, either one of the below, or or a mix of these, depending on the tag:
  - simply whether the tag exists in a list or not, (e.g., if `collage` is in the tag list, then it is a collage, and if missing, not a collage)
  - all boolean (e.g., `collage:[false,true]`), with missing tags treated as unknown
  - multivalued (e.g., `collage:[true,false,unknown,not_applicable]`, or `environment:[indoor,outdoor,outer_space,unknown]`)
  - something else?
- clean sample code examples for accessing tags, making this very easy for preprocessors/handlers to deal with in a consistent way
- Is this a one-shot, or if we have multiple preprocessors that need to tag, is there a way to expand the list of tags?
- ???

jaydeepsingh25:
Resolves https://github.com/Shared-Reality-Lab/IMAGE-server/issues/1015. 
PR should be merged after https://github.com/Shared-Reality-Lab/IMAGE-server/pull/1136

This pull request updates how image content is classified as a photograph across several preprocessing modules. The changes standardize the use of the `categories` dictionary and its `photo` key, replacing previous logic that relied on a single `category` string value. This improves consistency and reliability when determining whether to process image content as a photograph.

**Standardization of photograph classification:**

* Updated `depth-map-generator.py`, `segment.py`, `azure_api.py`, and `objdetect.py` to use the `categories["photo"]` boolean for photograph detection, replacing checks against the `"category"` string. [[1]](diffhunk://#diff-0d9e241a50655b8994698107d61c7856fcae136b7d78e1783f9ebfc63643ee1dL106-R107) [[2]](diffhunk://#diff-362ea5a581e2ed9a0979e81ccaf3a496eae17ac9ee476ae3670a047f8ab4ef14L164-R166) [[3]](diffhunk://#diff-728bd5f021ea835ee88307ae0e6b000a1598929d9e4cc943a28a186c60281c05L142-R143) [[4]](diffhunk://#diff-877c528b964e592108a185d96bee2879b0ae7de497a9bd30c62e6591c2d0b542L140-R142)

**Logic improvements:**

* Replaced string comparison and negation logic with direct boolean checks, reducing potential errors and making the code easier to maintain. [[1]](diffhunk://#diff-0d9e241a50655b8994698107d61c7856fcae136b7d78e1783f9ebfc63643ee1dL106-R107) [[2]](diffhunk://#diff-362ea5a581e2ed9a0979e81ccaf3a496eae17ac9ee476ae3670a047f8ab4ef14L164-R166)

Don't delete below this line.

---

## Required Information

## Coding/Commit Requirements

OR

### DIFF
--- a/preprocessors/depth-map-gen/depth-map-generator.py
+++ b/preprocessors/depth-map-gen/depth-map-generator.py
@@ -103,8 +103,8 @@ def depthgenerator():
     classifier_1 = "ca.mcgill.a11y.image.preprocessor.contentCategoriser"
     if classifier_1 in preprocess_output:
         classifier_1_output = preprocess_output[classifier_1]
-        classifier_1_label = classifier_1_output.get("category", "")
-        if classifier_1_label != "photograph":
+        categoriser_tags = classifier_1_output["categories"]
+        if not categoriser_tags["photo"]:
             logging.info("Not photograph content. Skipping...")
             return "", 204
     else:
--- a/preprocessors/graphic-tagger/azure_api.py
+++ b/preprocessors/graphic-tagger/azure_api.py
@@ -139,9 +139,8 @@ def categorise():
         if content_classifier in preprocess_output:
             content_classifier_output = \
                 preprocess_output[content_classifier]
-            content_label = \
-                content_classifier_output["category"]
-            if content_label == "photograph":
+            categoriser_tags = content_classifier_output["categories"]
+            if categoriser_tags["photo"]:
                 source = content["graphic"]
                 image_b64 = source.split(",")[1]
                 binary = base64.b64decode(image_b64)
--- a/preprocessors/mmsemseg/segment.py
+++ b/preprocessors/mmsemseg/segment.py
@@ -161,8 +161,9 @@ def segment():
     # if the architecture is modified appropriately
     if classifier_1 in preprocess_output:
         classifier_1_output = preprocess_output[classifier_1]
-        classifier_1_label = classifier_1_output["category"]
-        if classifier_1_label != "photograph":
+        categoriser_tags = classifier_1_output["categories"]
+
+        if not categoriser_tags["photo"]:
             logging.info("Not photograph content. Skipping...")
             return "", 204
         if classifier_2 in preprocess_output:
--- a/preprocessors/object-detection-azure/objdetect.py
+++ b/preprocessors/object-detection-azure/objdetect.py
@@ -137,9 +137,9 @@ def categorise():
         if content_classifier in preprocess_output:
             content_classifier_output = \
                 preprocess_output[content_classifier]
-            content_label = \
-                content_classifier_output["category"]
-            if content_label == "photograph":
+            content_tags = \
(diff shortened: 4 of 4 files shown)


jeffbl (approved):
Few items, which could be fixed later, or addressed now:
1. @gvzdv I think graphicTagger will be deprecated in favor of LLM tagging, correct? If so, mmsemseg can be simplified, since it has strange checks for this preprocessor, which I'd argue should be removed anyway.
2. If I'm reading correctly, if the tag "photo" doesn't exist at all (as opposed to it being `false`, then most of these preprocessors will not run. In the past, we debated whether these should run anyway, if the value is effectively "unknown", e.g., if results from the categoriser didn't show up at all. I'd be curious to get @gvzdv comment on this, since it is partially dependent on how good we think the categorisation is. Either way, we'll need to test to make sure we're striking a good balance between running when the results are useful, and not running when the results are actively bad. Further caveat that in the future we might be replacing YOLO and semseg with LLM as well, which could complicate things further.

Surprised there were only four preprocessors that look at the categorisation!

@jaydeepsingh25 are we sure there are no handlers that look at these? (I imagine tests would have broken if so...)

gvzdv (approved):
Regarding @jeffbl's comments:
1. `graphic-tagger` can (and probably should) be deprecated. We can always add these tags to the `content-categoriser` schema, but (if I remember correctly) we decided that they are not necessary anymore.
2. I believe we can keep things as they are for now, but when we get to updating object detection and semantic segmentation, we will need to have a deeper discussion (with a paper trail) about that. It will heavily depend on what results we will be getting for non-photo graphics.

`object-detection-azure` isn't deployed at all, right? I don't see it in the `docker-compose.yml`

Another preprocessor that has an optional dependency on the `content-categoriser` is `yolo`, but there's no tag checks in the current code + we are set to try replacing it in the following weeks, so I would not worry about it too much.

tl;dr: looks good to me, nothing should break

PR #1164, opened 2025-10-22
LLM-based object detector

### LINKED ISSUE #1147: New object detection preprocessor based on LLM
YOLO, our current object detector, is not great, even though it is very fast. It regularly makes mistakes on animals, etc. Goal is to replace YOLO with an LLM, to increase accuracy.

NOTE: Recommend keeping YOLO as an option, so make a new preprocessor targeting LLMs, and we'll keep YOLO as another option (that we aren't using in production) until we find it too annoying to maintain.

gvzdv:
Resolves #1147 

Brand new preprocessor that uses the vision LLM to detect objects and produce their coordinates and labels.

Notes:
- Right now, I prompted the model to focus on the major and most important objects. This can be changed in the prompt file.
- Confidence threshold is 0.9. Can be adjusted via environment variables.
- If no objects are found, the preprocessor returns 503 to avoid triggering components depending on it.
- I believe I adjusted all components that depend on the object detector (`svg-od-handler`, `object-grouping`, `photo-tactile-svg-handler`) to work with this preprocessor instead of the YOLO-based one — will appreciate it if you double-check.
- The original YOLO-based preprocessor is not deprecated and still runs, but its output isn't being used anywhere.

Tested manually on Unicorn. Needs more extensive testing. I expect some edge cases (mostly solvable via prompting), but it won't have the obvious YOLO problems.

---

## Required Information

## Coding/Commit Requirements

* [x] I have created a `README.md` file that describes what the component does and what it depends on (other microservices, ML models, etc.).

OR


handlers/photo-tactile-svg/tactile_svg.py
```diff
@@ -152,14 +152,17 @@ def handle():
     form = inflect.engine()
     caption = ""
 
-    if "ca.mcgill.a11y.image.preprocessor.objectDetection" in preprocessors\
```
jeffbl: why is this looking at the specific preprocessor, rather than just for object detection output schema data?

jeffbl (changes requested):
Excited to see output. Did you see significantly better results than with YOLO? The blocking issue with merge is that I'm unclear, if the llm object detector is using the same schema as the yolo od, why changes were necessary in later preprocessors/handlers, which I thought would be looking for data in a particular schema, not from a particular preprocessor. From what I can tell, if we wanted to run with YOLO again, we'd have to make changes to a number of preprocessors/handlers, which breaks the modular intent of the IMAGE stack? Let me know if easiest to jump on zoom to discuss, or if I'm missing something obvious.

handlers/photo-tactile-svg/tactile_svg.py
gvzdv: This is more of a system architecture question rather than a component question. It works like that in most preprocessors/handlers that depend on other components, and it's been like that long before I joined IMAGE... We need to discuss it before making changes, because (from what I understand) it may result in a large work item.

In this specific case, I could have just named the new preprocessor `ca.mcgill.a11y.image.preprocessor.objectDetection` and renamed the YOLO preprocessor to something different. Then, all the components depending on the object detector would automatically pick up the new one (in hindsight, it looks like a better strategy). But that would still leave the question about us hardcoding preprocessor names into other preprocessors and handlers.

Let's discuss this issue next week before/after the lab meeting?

gvzdv:
Changed 503 to 204 and added a little filter that removes underscores in labels ("traffic_light" -> "traffic light").

jeffbl (approved):

PR #1203, opened 2026-08-03
fix: support both qwen and gemma coordinate systems (#1167)

### LINKED ISSUE #1167: Select and validate the production vision-language model
New [Qwen vision model](https://qwen.ai/blog?id=99f0335c4ad9ff6153e517418d48535ab6d8afef&from=research.latest-advancements-list) is out, and it's (on paper) better than the 2.5 that we use now.

I would like to try it out, which means we'll need to set aside time to swap models on Pegasus and test the new one.
There are several model variations (thinking/non-thinking, various sizes). I am considering a non-thinking (for speed) non-quantized 8B model — let's see if we can fit it on our GPU, especially if we can deprecate YOLO. Alternatives would be an 8-bit 8B model or an FP16 4B model.

MansiDhanania:
Contributes to [#1167](https://github.com/Shared-Reality-Lab/IMAGE-server/issues/1167)

### Changes made
As part of the investigation in [#1167](https://github.com/Shared-Reality-Lab/IMAGE-server/issues/1167) (selecting/validating the production vision-language model), the [object-detection-llm](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/preprocessors/object-detection-llm/object-detection-llm.py) and [multistage-diagram-segmentation](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py) preprocessors are now made compatible with both Qwen and Gemma, since the two models use different bounding box key names and coordinate systems. This does not select a final model, it only accommodates both coordinate systems so that both can be tested and compared going forward.

Both preprocessors now detect the active model and adapt the labels and coordinate systems accordingly:
- Qwen: `bbox_2d` key, `[x1, y1, x2, y2]` order
- Gemma: `box_2d` key, `[y1, x1, y2, x2]` order

This required:
- Parameterizing the prompt templates in [`prompts.py`](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/utils/llm/prompts.py) to use the correct key/order per model
- Updating the loaded JSON schemas so the LLM is constrained to output the correct key
- Adding a conversion step in each preprocessor to translate the coordinate output back into the format the rest of the pipeline expects (`[x1, y1, x2, y2]`).

### How I tested this
- [x] Verified the prompt templates format for both key/order values (Qwen and Gemma) using a test script
- [x] Override on Unicorn

---

## Required Information

## Coding/Commit Requirements

OR


preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -37,6 +37,19 @@
...
+# Get model name from environment variable
+LLM_MODEL = os.environ.get('LLM_MODEL', '').lower()
+MODEL_NAME = "gemma" if "gemma" in LLM_MODEL else "qwen"
+logging.debug(f"Using LLM model: {LLM_MODEL}, interpreted as {MODEL_NAME}")
+
+# Set bounding box key based on model
+BBOX_KEY = "box_2d" if MODEL_NAME == "gemma" else "bbox_2d"
+
+# Set coordinate order based on model
+bbox_order = "[y1, x1, y2, x2]" if MODEL_NAME == "gemma" else "[x1, y1, x2, y2]"
```
aidanbruneel: Line 49 exceeds 79 characters.

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -37,6 +37,19 @@
...
+
+FORMATTED_PROMPT = OBJECT_DETECTION_PROMPT.format(bbox_key=BBOX_KEY, bbox_order=bbox_order)
```
aidanbruneel: Line 51 exceeds 79 characters.

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -46,6 +59,16 @@
 with open(BBOX_SCHEMA, 'r') as f:
     BBOX_RESPONSE_SCHEMA = json.load(f)
 
+# Swap the bounding box key to match the active model's convention
+# Note: If the schema is loaded twice within a run, this might give a key error => Needs testing
```
aidanbruneel: Line 63 exceeds 79 characters.

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -68,30 +95,31 @@ def normalize_bbox(bbox, width, height):
...
 
     Returns:
         list: Processed objects with computed properties
     """
     processed = []
-    for idx, item in enumerate(qwen_output):
+    for idx, item in enumerate(llm_output):
         # Normalize bounding box
-        x1, y1, x2, y2 = normalize_bbox(item["bbox_2d"], width, height)
+        x1, y1, x2, y2 = normalize_bbox(item[BBOX_KEY], width, height, model_name)
```
aidanbruneel: Line 122 exceeds 79 characters.

aidanbruneel (changes requested):
`object-detection-llm.py` is just failing the PEP 8 check.

I will let you resolve this and then give a more in-depth review once checks are passing.

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
```diff
@@ -63,6 +89,27 @@
     logging.error(f"Failed to initialize clients: {e}")
     sys.exit(1)
 
+# Convert bounding boxes to expected format for SAM
+
+
```
aidanbruneel: Delete whitespace.

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
```diff
@@ -31,11 +31,27 @@
...
 import json
+import os
 
 configure_logging()
 
 app = Flask(__name__)
 
+# Get model name from environment variable
+LLM_MODEL = os.environ.get('LLM_MODEL', '').lower()
+MODEL_NAME = "gemma" if "gemma" in LLM_MODEL else "qwen"
```
aidanbruneel: This will work right now, but what happens when we have additional LLMs?

I imagined a sort of key-value dictionary where you could specify a model (and even version), and then utilize the correct co-ordinate system paradigm. Right now, that is `gemma` and `qwen`. In the future, it will be more.

Also, LLM stands for "Large Language Model", so this variable name is basically "LARGE_LANGUAGE_MODEL_MODEL". I can see you are using the string `LLM_MODEL` inside `os.environ.get()`, so I imagine there is some convention behind this I am just not familiar with.

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -55,11 +83,15 @@
     sys.exit(1)
 
 
-def normalize_bbox(bbox, width, height):
+def normalize_bbox(bbox, width, height, model_name):
     """
-    Normalize bounding box coordinates to [0,1] range
+    Normalize bounding box coordinates to [0,1] range.
+    Handles differing coordinate orders between Qwen and Gemma.
```
aidanbruneel: Again, let's frame this as being designed for any LLM, and specify our two currently supported LLMs (`qwen` and `gemma`) in a key-value dictionary or something to that effect.

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -55,11 +83,15 @@
...
     """
-    x1, y1, x2, y2 = bbox
+    if model_name == "gemma":
+        y1, x1, y2, x2 = bbox
+    else:  # Default set to Qwen's order
+        x1, y1, x2, y2 = bbox
```
aidanbruneel: This should be compared against a config/dictionary, not assume that any value other than `"gemma"` must be `"qwen"`.

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -68,30 +100,33 @@ def normalize_bbox(bbox, width, height):
...
     - Calculates geometric properties (area, centroid)
     - Filters objects by confidence threshold
 
     Args:
-        qwen_output (list): Qwen detection output with bbox_2d and label
+        llm_output (list): LLM detection output with bbox_2d and label
         width (int): Image width in pixels for normalization
         height (int): Image height in pixels for normalization
         threshold (float): Minimum confidence score (0-1)
+        model_name (str): Name of the LLM model used
```
aidanbruneel: Imagine that `model_name` could, in the future, be any LLM model name. ChatGPT, Claude, Qwen, Gemma, anything.
Right now you only need to implement the co-ordinate system specifications for Qwen and Gemma, but it needs to be designed so that we can just append more model support later.

aidanbruneel (changes requested):
Overall the functionality is here but it is written to be too limited/specific to the two models you were investigating. Since we are now introducing the possibility of more than one model, we should do so in a model agnostic way.

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
MansiDhanania: Re flexibility across models, I agree. However, using a different model will incur other problems with regards to the normalization and post-processing of coordinates (gemma and qwen both use the same normalization). So that would require a "larger" accommodation. I could add that in under a separate fix.

Re LLM_MODEL, that is how it's set in the env file.

aidanbruneel: @MansiDhanania I suppose what I am saying is that for the parts of an LLM's "co-ordinate specification" that we are currently dealing with, I do not want `MODEL_NAME = "gemma" if "gemma" in LLM_MODEL else "qwen"`  sort of design patterns with a binary `if ... else ...` statement that depends on whether or not the `MODEL_NAME` is `"gemma"`. What if the `MODEL_NAME` is `""` or `"gemmah"` or `"gwen"` as a typo or type problem?

Just as an example,
1. Loosely define model configs
```
from dataclasses import dataclass


@dataclass(frozen=True)
class ModelConfig:
    identifiers: tuple[str, ...]
    bbox_order: tuple[str, str, str, str]


MODEL_CONFIGS: dict[str, ModelConfig] = {
    "gemma": ModelConfig(
        identifiers=("gemma",),
        bbox_order=("y1", "x1", "y2", "x2"),
    ),
    "qwen": ModelConfig(
        identifiers=("qwen",),
        bbox_order=("x1", "y1", "x2", "y2"),
    ),
}
```
2. Resolve the model family name
```
def resolve_model_name(llm_model: str) -> str:
    normalized_name = llm_model.casefold()

    for model_name, config in MODEL_CONFIGS.items():
        if any(identifier in normalized_name for identifier in config.identifiers):
            return model_name

    raise ValueError(f"Unsupported LLM model: {llm_model}")
```
- Sidenote: I wonder if it might be better to have a specific config for each model version in case qwen or gemma or some other LLM changes their convention with a new version

3. Now you can replace L42 with something like:
```
model_name = resolve_model_name(LLM_MODEL)
```

This would be a more robust architecture and does not require us to speculate on how to define configs in the future. It's just the `if ... else ...` pattern I want to avoid.

@jeffbl Do you agree that this is a more ideal design pattern?

aidanbruneel: TL;DR: a repeated `if ... else ...` design pattern where any value other than `"gemma"` is automatically assumed to be the same as the value `"qwen"` is tech debt not worth owing.

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
```diff
@@ -31,11 +31,27 @@
...
+logging.debug(f"Using LLM model: {LLM_MODEL}, interpreted as {MODEL_NAME}")
+
+# Set bounding box key based on model
+BBOX_KEY = "box_2d" if MODEL_NAME == "gemma" else "bbox_2d"
```
aidanbruneel: This is the same sort of `if ... else ...` design pattern that assumes `"qwen"` if not `"gemma"` that I want to avoid.

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
```diff
@@ -63,6 +89,27 @@
...
+    for item in bboxes_data:
+        coords = item[bbox_key]
+        if model_name == "gemma":
+            y1, x1, y2, x2 = coords
+        else:
+            x1, y1, x2, y2 = coords
+        normalized.append({
+            "bbox_2d": [x1, y1, x2, y2],
+            "label": item["label"]
+        })
```
aidanbruneel: `if ... else ...` design pattern that only depends on whether or not the `model_name` is `gemma`.

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -37,6 +37,24 @@
 
 CONF_THRESHOLD = float(os.environ.get('CONF_THRESHOLD', '0.9'))
 
+# Get model name from environment variable
+LLM_MODEL = os.environ.get('LLM_MODEL', '').lower()
+MODEL_NAME = "gemma" if "gemma" in LLM_MODEL else "qwen"
```
aidanbruneel: `if ... else ...` design pattern

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -37,6 +37,24 @@
...
+logging.debug(f"Using LLM model: {LLM_MODEL}, interpreted as {MODEL_NAME}")
+
+# Set bounding box key based on model
+BBOX_KEY = "box_2d" if MODEL_NAME == "gemma" else "bbox_2d"
```
aidanbruneel: `if ... else ...` design pattern

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -37,6 +37,24 @@
...
+
+# Set coordinate order based on model
+bbox_order = (
+    "[y1, x1, y2, x2]" if MODEL_NAME == "gemma" else "[x1, y1, x2, y2]"
```
aidanbruneel: `if ... else ...` design pattern

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -46,6 +64,16 @@
 with open(BBOX_SCHEMA, 'r') as f:
     BBOX_RESPONSE_SCHEMA = json.load(f)
 
+# Swap the bounding box key to match the active model's convention
+# Note: if schema is loaded twice, this might raise a key error
+prop = BBOX_RESPONSE_SCHEMA["items"]["properties"]
+prop[BBOX_KEY] = prop.pop("bbox_2d")
+prop[BBOX_KEY]["description"] = f"Bounding box coordinates {bbox_order}."
+BBOX_RESPONSE_SCHEMA["items"]["required"] = [
+    BBOX_KEY if k == "bbox_2d" else k
```
aidanbruneel, line 85: This one might be okay if the only two _possible_ values of `k` in the future are `bbox_2d` or `box_2d`. Otherwise, it is an `if ... else ...` design pattern that would need to be restructured anyway in the future whenever we add another LLM compatibility.

aidanbruneel (commented):
Just identifying other locations with what I think is an undesirable and repeated `if ... else ...` design pattern that assumes `"qwen"` if not `"gemma"`. I always think to myself "am I repeating the same sort of code repeatedly?", and if so, it typically belongs in a function or needs to be leaner.

utils/llm/prompts.py
```diff
@@ -17,23 +17,23 @@
 Step 2:
 Give the bounding boxes for the objects determined in the first step.
 Output a only JSON list of bounding boxes where each entry contains:
-- the pixel coordinates of a 2D bounding box in the key "bbox_2d",
+- the pixel coordinates of a 2D bounding box in the key "{bbox_key}",
```
aidanbruneel: I also just noticed this statement about it being "pixel coordinates" might now be inaccurate if the values are normalized 0-1000.

preprocessors/object-detection-llm/object-detection-llm.py
aidanbruneel: Feel free to resolve this if I am wrong.

preprocessors/object-detection-llm/object-detection-llm.py
aidanbruneel:
```suggestion
    Handles differing coordinate orders between LLMs.
```

utils/llm/prompts.py
aidanbruneel:
```suggestion
- the normalized coordinates (0-1000) of a 2D bounding box in the key "{bbox_key}",
```

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
MansiDhanania: This was added to satisfy the PEP 8 check

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
jeffbl: @aidanbruneel agree better for future.

MansiDhanania: Apologies, I saw this comment only after the latest edits, but the functionality is similar to my latest commit. Hope that satisfies your concern.

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
aidanbruneel: My bad!

utils/llm/coordinate_convention.py
```diff
@@ -0,0 +1,35 @@
...
+    """Determine model family from the LLM_MODEL env value."""
+    llm_model_env = (llm_model_env or "").lower()
+    for family in MODEL_BBOX_FORMATS:
+        if family in llm_model_env:
+            return family
+    return DEFAULT_FAMILY
+
+def get_bbox_format(model_family):
+    """Return the bbox config dict for a given model family."""
+    return MODEL_BBOX_FORMATS.get(model_family, MODEL_BBOX_FORMATS[DEFAULT_FAMILY])
```
aidanbruneel: If the value of `llm_model_env` is a typo, unsupported model, or empty/null/undefined, this will return `"qwen"`.

Also if you are going to force `llm_model_env` to `.lower()`, you might as well use a better local variable name like `normalized_model`. `llm_model_env` is what the function parameter is basically requesting as input, but we are locally handling it to check for the family. It doesn't matter that much though so it's up to you.

Also, I prefer to type my parameters when it is something coming from an outside source (`.env`) like this.

```suggestion
def get_model_family(llm_model_env: str | None) -> str:
    """Determine model family from the LLM_MODEL env value."""
    normalized_model = (llm_model_env or "").lower()

    for family in MODEL_BBOX_FORMATS:
        if family in normalized_model:
            return family

    raise ValueError(
        f"Unsupported or unrecognized LLM model: {llm_model_env!r}"
    )


def get_bbox_format(model_family: str) -> dict:
    """Return the bbox config dict for a given model family."""
    return MODEL_BBOX_FORMATS[model_family]
```

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
```diff
@@ -28,14 +28,33 @@
...
 import json
+import os
 
 configure_logging()
 
 app = Flask(__name__)
 
+# Get model name from environment variable
+LLM_MODEL = os.environ.get('LLM_MODEL', '').lower()
+MODEL_NAME = get_model_family(LLM_MODEL)
```
aidanbruneel: We now have the notion of a Model Family rather than a Model Name.

```suggestion
MODEL_FAMILY = get_model_family(LLM_MODEL)
```

You will need to go and update anywhere this constant is referenced as well.

preprocessors/multistage-diagram-segmentation/multistage-diagram-segmentation.py
```diff
@@ -28,14 +28,33 @@
...
+
+# Get bounding box format based on model
+BBOX_FORMAT = get_bbox_format(MODEL_NAME)
+
+# Set bounding box key based on model
+BBOX_KEY = BBOX_FORMAT["bbox_key"]
+
+# Set coordinate order based on model
+COORD_ORDER = BBOX_FORMAT["coord_order"]
+bbox_order = f"[{', '.join(COORD_ORDER)}]"
```
aidanbruneel: This is a formatted string that takes in a constant, so it is also a constant.

```suggestion
BBOX_ORDER = f"[{', '.join(COORD_ORDER)}]"
```

You will need to go and update anywhere this constant is referenced as well.

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -113,7 +152,7 @@ def process_objects(qwen_output, width, height, threshold):
         processed.append(obj)
 
     logging.debug(
-        f"Processed {len(qwen_output)} objects from Qwen output"
+        f"Processed {len(llm_output)} objects from LLM output"
     )
     return processed
```
aidanbruneel: Pass the complete configuration into `process_objects()` instead of an incomplete list of arguments and then locally reading the global `BBOX_KEY`.

```suggestion
def process_objects(
    llm_output: list[dict[str, Any]],
    width: int,
    height: int,
    threshold: float,
    bbox_format: Mapping[str, Any],
) -> list[dict[str, Any]]:
    bbox_key = bbox_format["bbox_key"]
    coord_order = bbox_format["coord_order"]
    """
    Transform model-specific object detections into the IMAGE object schema.

    Uses ``bbox_format`` to identify the bounding-box key and coordinate order,
    then normalizes coordinates and calculates each object's area and centroid.

    Args:
        llm_output: Detected objects returned by the LLM.
        width: Image width in pixels.
        height: Image height in pixels.
        threshold: Confidence value assigned to each object.
        bbox_format: Bounding-box configuration containing ``bbox_key`` and
            ``coord_order``.

    Returns:
        Objects formatted according to the IMAGE schema.
    """
    processed = []

    for idx, item in enumerate(llm_output):
        x1, y1, x2, y2 = normalize_bbox(
            item[bbox_key],
            width,
            height,
            coord_order,
        )

        processed.append({
            "ID": idx,
            "type": item["label"].replace("_", " "),
            "confidence": threshold,
            "dimensions": [x1, y1, x2, y2],
            "area": (x2 - x1) * (y2 - y1),
            "centroid": [
                (x1 + x2) / 2,
                (y1 + y2) / 2,
            ],
        })

    return processed
```

You will also need to add the following imports up in the top of the file:
```
from collections.abc import Mapping
from typing import Any
```

You will also need to go and change where this function is called to:
```
processed_objects = process_objects(
    llm_output,
    width,
    height,
    CONF_THRESHOLD,
    BBOX_FORMAT,
)
```

I also broke out the function's parameters into separate lines; 6 parameters is harder to read inline.

I also leaned out the function header comment you included as it was much more verbose than any of our existing headers; perhaps it is justified to give some clarification to the complete configuration used as input, but comments in code like this need to be maintained as well, so having really verbose comments creates more opportunities for them to drift from the truth.

utils/llm/prompts.py
aidanbruneel: This still needs to be fixed.

preprocessors/object-detection-llm/object-detection-llm.py
aidanbruneel: This still needs to be fixed.

preprocessors/object-detection-llm/object-detection-llm.py
aidanbruneel: Respond to this or resolve it please.

aidanbruneel (changes requested):
Comment cleanup, `process_objects()` improvement, style improvements, handling expected string value edge cases.

@MansiDhanania Do not just accept my suggestions and then request a review again, as some of my suggestions, if implemented, would involve repairing lines that reference them.

preprocessors/object-detection-llm/object-detection-llm.py
MansiDhanania: This code block mutates the loaded JSON schema so the LLM is constrained to output the correct bbox key for the active model (from coordinate_convention.py), instead of always requiring bbox_2d. Since get_model_family now exits the process with an error for any unsupported model before this code is reached, this mutation only ever runs with a valid bbox_key.

utils/llm/coordinate_convention.py
```diff
@@ -0,0 +1,36 @@
...
+
+"""
+LLM-specific bounding box format configuration.
+Add new model families here as they're supported.
+"""
+MODEL_BBOX_FORMATS = {
+    "qwen": {
+        "bbox_key": "bbox_2d",
+        "coord_order": ("x1", "y1", "x2", "y2"),
+    },
```
aidanbruneel, line 16: It should be noted that Qwen 2.5, which is currently stated to be what the `object-detection-llm` preprocessor is tested on, uses absolute coordinates based on the resized image dimensions, whereas Qwen 3 VL returned to normalized 0–1000 coordinates. Qwen 3.5 follows the Qwen 3 VL normalized convention.

[This](https://github.com/QwenLM/Qwen3-VL/issues/866?utm_source=chatgpt.com) is where the co-ordinate system change happened for Qwen models.

Right now `get_model_family()` will happily classify both `qwen2.5...` and `qwen3...` as `"qwen"`, after which `normalize_bbox()` assumes 0–1000. This calls to attention that really, the configuration of co-ordinate systems should depend on the _specific_ model version being used, rather than handling all Qwen models under the normalized convention assumption.

The thing is, this won't practically matter unless one of the model families (e.g., Qwen, Gemma, GPT, Claude) that we use in the future undergoes any co-ordinate system change, or if we want to load up Qwen 2.5 for some reason, or if an open-source contributor/user wants to fork the IMAGE-server and use their own models (they would need to refactor this code for that to be possible).

After some deliberation, I am not going to mandate that you redesign this to handle specific model configurations, but I'd consider it [tech debt](https://en.wikipedia.org/wiki/Technical_debt).

preprocessors/object-detection-llm/object-detection-llm.py
```diff
@@ -68,30 +108,34 @@ def normalize_bbox(bbox, width, height):
     ]
 
 
-def process_objects(qwen_output, width, height, threshold):
+def process_objects(llm_output: list[dict[str, Any]], 
+                    width: int, 
+                    height: int, 
+                    threshold: float, 
+                    bbox_format: Mapping[str, Any]
+                    ) -> list[dict[str, Any]]:
```
aidanbruneel: Trailing whitespace on these lines. I believe that would cause our `flake8` workflow to fail under `W291 trailing whitespace`.

```suggestion
def process_objects(llm_output: list[dict[str, Any]],
                    width: int,
                    height: int,
                    threshold: float,
                    bbox_format: Mapping[str, Any]
                    ) -> list[dict[str, Any]]:
```

utils/llm/coordinate_convention.py
```diff
@@ -0,0 +1,36 @@
...
+        if family in llm_model_env:
+            return family
+    raise ValueError(
+        f"Unsupported LLM model: '{llm_model_env}'. "
+        f"Supported families: {list(MODEL_BBOX_FORMATS.keys())}"
+    )
+
+def get_bbox_format(model_family: str) -> dict:
+    """Return the bbox config dict for a given model family."""
+    return MODEL_BBOX_FORMATS[model_family]
```
aidanbruneel: No newline at end of file.


```suggestion
    return MODEL_BBOX_FORMATS[model_family]

```

aidanbruneel (changes requested):
I noticed a few more formatting issues and would also like to point out that the [`preprocessors/object-detection-llm/README.md`](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/preprocessors/object-detection-llm#:~:text=Note%3A%20This%20preprocessor%20is%20developed%20to%20be%20used%20with%20Qwen%20VL%20family%20of%20models%20(tested%20on%20Qwen%202.5%20VL)%20due%20to%20their%20ability%20to%20correctly%20identify%20object%20bounding%20boxes.) needs to be updated to reflect changes.

It's up to you whether you want to address the specific-model comment I made; I would just consider it tech debt if we are using this family-of-models approach. I imagine with more models supported down the line, there will be even more examples of model versions that diverge on their co-ordinate system within the same family.

utils/llm/coordinate_convention.py
MansiDhanania: Noted and agreed.

MansiDhanania: Updated the READMEs to reflect this.

aidanbruneel (approved):
Looks good to me. :)

PR #1213, opened 2026-08-17
Feature: Per-preprocessor endpoint override (#973)

### LINKED ISSUE #973: Enable cloud LLMs to be used by preprocessors
the ollama.env file allows you to specify the local ollama / open-webui URL for making LLM queries. Goal of this work item is to test and document, and make code changes if necessary, to allow the use of cloud services as well, for a few reasons:

1. Can use larger models than we can fit on our local GPUs
2. Can run server without a GPU at all (or with wimpy GPUs like on our test server)
3. Have a backup in case our production server goes down
4. Have the ability to use local GPU until load increases, then start using cloud to keep query latency low

Note:
1. A single variable for all preprocessors (as currently implemented) probably doesn't make sense, since we would likely want to run smaller models locally, and larger ones in the cloud
2. This cannot go to production until our TOS is updated to allow data to be sent to cloud
3. Documenting what is tested, with examples for different cloud services, would be useful to others setting up IMAGE servers
4. Suggest that @shahdyousefak documentation for bringing up a new IMAGE server default to this, since it would mean a faster path to getting something working

MansiDhanania:
Related to [#973](https://github.com/Shared-Reality-Lab/IMAGE-server/issues/973)

## Summary
Extended and documented the per-preprocessor LLM configuration override pattern. This includes a shared default endpoint for all LLM-related preprocessors, with the ability for any individual preprocessor to override the endpoint/model independently.

## Initial Tests and Results
@Miliya-Ai's tests showed that the specifications in [`config/llm.env`](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/config) work with any OpenAI-compatible cloud endpoint called via [`client.py`](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/utils/llm/client.py). The endpoint is used for all preprocessors requiring this service. 

## Changes made
In order to allow for different preprocessors, orchestrators or handlers to call a different endpoint, the following changes were incorporated:

The [docker-compose.yml](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/docker-compose.yml) file was extended to include the paths to both a global and local `.env` file:
```yaml
env_file:
  - path: ./config/llm.env
    required: true
  - path: ./preprocessors/<name>/config/<name>.env
    required: false
```

The global [llm.env](https://github.com/Shared-Reality-Lab/IMAGE-server/tree/main/config) is always loaded first. A preprocessor-specific file, if present, loads second and overrides only the variables it sets for that one preprocessor. If the override file doesn't exist, the preprocessor falls back to the shared default.

This addresses the issue as there now exists:
- A shared default endpoint/model for preprocessors using the main LLM
- A service-level override when needed
- Setup documented in [DEPLOYMENT.md](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/cloud-llms/DEPLOYMENT.md), [config/Readme.md](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/cloud-llms/config/Readme.md) and an example of the same in [preprocessors/object-detection-llm/config/Readme.md](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/cloud-llms/preprocessors/object-detection-llm/config/Readme.md).

## How I tested this

- [x] Tested on unicorn: created `preprocessors/object-detection-llm/config/object-detection-llm.env` calling Qwen3.8 via Mallory, while leaving `config/llm.env` pointed at the default Gemma 4 Pegasus endpoint. Recreated only the `object-detection-llm` container via override and queried an image.
- [x] Confirmed via logs that:
- `object-detection-llm` picked up the Qwen override.
- Other preprocessors (e.g., `content-categoriser`) continued to call Gemma.

---
## Required Information

## Coding/Commit Requirements

OR


DEPLOYMENT.md
```diff
@@ -199,9 +199,11 @@ networks:
 ```
 
 # Connecting to, or running your own, visual LLM
-Since multiple IMAGE preprocessors us a visual LLM (currently qwen), you need to either specify a cloud endpoint, or run your own LLM locally on your server.
+Since multiple IMAGE preprocessors use a visual LLM (currently gemma), you need to either specify a cloud endpoint, or run your own LLM locally on your server.
```
jeffbl: Since multiple IMAGE preprocessors use a visual LLM, you need to specify an endpoint. We have run with a local LLM served by vLLM or ollama, but cloud endpoints also work. As of August, 2026, we have tested with

.gitignore
```diff
@@ -1,21 +1,9 @@
...
-config/sendimagereq.env
-config/slack-webhook-preprocessors.env
-config/imageup.env
-config/gemini.env
 *.sh
 docker-compose.override.yml
+# Ignore all files in config/ except documentation and example files
+config/*
+!config/Readme.md
+!config/*.example.env
```
jeffbl: I believe .env.example is the more common way of doing this? That way if you make a copy, they also sort side-by-side.

docker-compose.yml
```diff
@@ -502,7 +523,10 @@ services:
     image: ghcr.io/shared-reality-lab/image-preprocessor-ocr-clouds:${REGISTRY_TAG}
     restart: "no"
     env_file:
-      - ./config/apis-and-selection.env
+      - path: ./config/apis-and-selection.env
```
jeffbl: This name makes no sense under the new naming system? If we don't have a general name, should we just go with one specifically for that preprocessor, with no fallback? At minimum, the fallback should be renamed (either around OCR, or maybe just around azure, if that is what it is really for?) @aidanbruneel do you think the configuration of the OCR tool to use should be in the .env or in the docker-compose directly? Last, OCR is not used in production, and is probably a candidate for deprecation at this point?

jeffbl:
Looks good, and I love that we have a more disciplined system. Couple things below. In addition, suggest:
1. in README, remove the example keys and such, and just make example files with endpoints we have tested already filled in, so people can see exactly what we have working (without our keys, of course!)
2. the text between DEPLOYMENT.md and the README.md seems repetitive. Could be postponed, but this should probably be tightened up in the future.

jeffbl (changes requested):

jeffbl:
Also, if this resolves #973, I don't see it linked to this item?

docker-compose.yml
aidanbruneel: > I agree! [express-common.env](https://github.com/Shared-Reality-Lab/IMAGE-server/blob/main/config/express-common.env) also only specifies a MAX_BODY field, so it's not really most suitable as a .env file.

Definitely transition this to Docker Compose.

aidanbruneel: @MansiDhanania I discussed this with @jeffbl so that we could keep the scope of this PR contained.

For OCR specifically, we recommend:

- Rename `apis-and-selection.env` to something tied to the preprocessor, rather than creating a general fallback name.
- Do not rename all existing configuration files as part of this PR.
- Keep the OCR provider credentials in that env file, since they are secrets.
- Move `CLOUD_SERVICE` and `MAX_BODY` into the Docker Compose configuration.
- Remove the optional `ocr.env` fallback because OCR does not have a shared default that needs a per-preprocessor override.
- Update the relevant README to match the resulting configuration.

Whether OCR should be deprecated is a separate repository-level decision and should not expand or block this PR.

MansiDhanania: Made the above stated changes in my latest commit. Ready for review!

Note that once these have been merged to main, we would have to change the .env file name on both unicorn and pegasus.

docker-compose.yml
```diff
@@ -502,14 +515,16 @@ services:
...
+      - path: ./config/cloud-ocr.env
+        required: true
     labels:
       ca.mcgill.a11y.image.preprocessor: 4
       ca.mcgill.a11y.image.port: 5000
       ca.mcgill.a11y.image.cacheTimeout: 3600
       ca.mcgill.a11y.image.required_dependencies: ""
       ca.mcgill.a11y.image.optional_dependencies: "object-detection-llm"
     environment:
+      - CLOUD_SERVICE="AZURE_READ"
```
aidanbruneel: Should this be `CLOUD_SERVICE=AZURE_READ` instead? In the array form of Compose `environment`, this is a `KEY=VALUE` string rather than shell assignment syntax, so I believe the quotes may become part of the value. `ocr.py` compares this directly against `"AZURE_READ"`, so if the container receives the quotes the provider selection will not match. Could you confirm what `docker compose config` / the container environment resolves this to on unicorn?

docker-compose.yml
```diff
@@ -547,14 +562,13 @@ services:
       ca.mcgill.a11y.image.required_dependencies: "ocr-clouds-preprocessor,object-detection"
       ca.mcgill.a11y.image.optional_dependencies: ""
     environment:
+      - CLOUD_SERVICE="AZURE_READ"
```
aidanbruneel: Similar to this comment:
https://github.com/Shared-Reality-Lab/IMAGE-server/pull/1213/changes#r3905355836

Does `ocr-handler` actually consume `CLOUD_SERVICE`? I can find it being read by `preprocessors/ocr/ocr.py`, but not by the handler. Since the provider selection belongs to `ocr-clouds-preprocessor`, I think this may have been added here unintentionally.

config/express-common.env
aidanbruneel: Since this PR removes `config/express-common.env` and moves `MAX_BODY` directly into the Compose service definitions, should this example and the explanation below it be updated as part of the same change? Otherwise the `orchestrator/README.md` will point new deployments to a file that no longer exists.

.gitignore
```diff
@@ -1,21 +1,9 @@
...
-config/make_request.env
-config/pp_daily_test.env
-config/sendimagereq.env
-config/slack-webhook-preprocessors.env
-config/imageup.env
-config/gemini.env
 *.sh
 docker-compose.override.yml
+# Ignore all files in config/ except documentation and example files
+config/*
```
aidanbruneel: Is ignoring the entire `config/` directory intentional? There are also tracked source files under `config/` (`logging_utils.py`, `process_image.py`, etc.). Existing tracked files are unaffected, but any new source file added there would be silently ignored. Would it be safer to ignore the runtime `.env`/secret files specifically instead?

DEPLOYMENT.md
```diff
@@ -217,8 +219,30 @@ You must specifiy services such as:
...
+
+To set one up, create a `.env` file within the existing `config/` directory using the naming convention `<preprocessor-name>.env`. Set `LLM_API_KEY`, `LLM_URL`, and `LLM_MODEL`. Add this path to the `docker-compose.yml` file's `env_file` entry for that service (if it does not exist) and set the `required` flag to `False`.
+
+**If you override any one of these three, set all three**: they describe a single endpoint together, and partially overriding them will send incorrect credentials to the new endpoint and fail the request.
+
+Preprocessors currently supporting this override pattern (check 
+`docker-compose.yml` for the current list, as this may grow): 
+`autour-preprocessor`, `content-categoriser`, `graphic-caption`, `text-followup`, `object-detection-llm`, `multistage-diagram-segmentation`.
+
+The same is applicable for any of the preprocessors. Check the preprocessor-specific `.env.example` file for setting env variables.
```
aidanbruneel: I think this list is mixing the generic two-env-file pattern with the specific LLM override described by this section. `autour-preprocessor` has a `maps.env` override rather than an `llm.env` override. Also, below we refer to preprocessor-specific `.env.example` files, but those don't appear to exist. Should this section point users to `llm.env.example` for the variables and list only the LLM consumers here?

aidanbruneel (commented):
@MansiDhanania I would like responses to these questions/concerns before approving.

docker-compose.yml
MansiDhanania: Yes you're right! Made the edit.

docker-compose.yml
MansiDhanania: You're right! The handler does not fetch it. Edited!

config/express-common.env
MansiDhanania: Updated orchestrator/README.md

.gitignore
MansiDhanania: Missed that! Edited

DEPLOYMENT.md
MansiDhanania: Preprocessor-specific `.env.example` files include `cloud-ocr.env.example`, `maps.env.example`, etc.

There is a more detailed explanation for the LLM configuration within `config/Readme.md`. I wanted to highlight that such an override could be created for any of the preprocessors with their own variables.

preprocessors/ocr/README.md
```diff
@@ -23,16 +23,10 @@ The desired cloud service is determined by an environment variable called `CLOUD
 
 The path to an environment file containing this variable should be provided in the `docker-compose.yml`, right in the `env_file` field of the `ocr-clouds-preprocessor` service.
 ## Environment setup
-The environment file (apis-and-selection.env) should contain the desired cloud service to be used, and the corresponding api keys.
+The environment file (cloud-ocr.env) should contain the desired cloud service to be used, and the corresponding api keys.
```
aidanbruneel: This still seems inconsistent with the configuration above. `cloud-ocr.env` now contains the provider credentials, while `CLOUD_SERVICE` is selected directly in `docker-compose.yml`. Could we update this sentence accordingly?

DEPLOYMENT.md
```diff
@@ -217,8 +219,30 @@ You must specifiy services such as:
...
+
+To set one up, create a `.env` file within the existing `config/` directory using the naming convention `<preprocessor-name>.env`. Set `LLM_API_KEY`, `LLM_URL`, and `LLM_MODEL`. Add this path to the `docker-compose.yml` file's `env_file` entry for that service (if it does not exist) and set the `required` flag to `False`.
+
+**If you override any one of these three, set all three**: they describe a single endpoint together, and partially overriding them will send incorrect credentials to the new endpoint and fail the request.
+
+Preprocessors currently supporting this override pattern (check 
+`docker-compose.yml` for the current list, as this may grow): 
+`content-categoriser`, `graphic-caption`, `text-followup`, `object-detection-llm`, `multistage-diagram-segmentation`.
+
+The override functionality is applicable for any of the preprocessors. Check the preprocessor-specific `.env.example` file for setting env variables.
```
aidanbruneel: I see what you mean from your reply. I think this sentence is still a little ambiguous in this LLM-specific section, though. `maps.env.example` and `cloud-ocr.env.example` document shared/base configurations rather than examples of the `<preprocessor-name>.env` override described above. Could we say that the same layered `env_file` pattern can be used for other service configurations, while pointing LLM overrides specifically to `llm.env.example`?

aidanbruneel (commented):
Some further questions/concerns and also the PR description still shows the override at `./preprocessors/<name>/config/<name>.env`, while the final implementation uses `./config/<preprocessor-name>.env`. It also links to `preprocessors/object-detection-llm/config/Readme.md`, which doesn't appear to exist on this branch. Could you update the description to match the final implementation?

aidanbruneel (approved):
Looks good to me :)

